PR #137, opened 2023-09-26
Reduce lock contention by backfilling new columns in batches

### LINKED ISSUE #127: Perform backfill operations in batches
The current implementation of backfilling for those operations that create a new column is naive:

https://github.com/xataio/pgroll/blob/5cbb9da7a698c2bb766949c9e326b8f0fc228c09/pkg/migrations/op_add_column.go#L223-L233

This operation should be done in batches to reduce lock contention.

andrew-farries:
The previous approach to backfilling new columns with values from existing columns applied one `UPDATE` statement to the whole table, taking row locks on the entire table in one long-running transaction.

This PR changes backfills to work in batches, copying 1000 records at a time, each in separate short transaction. This reduces contention for row locks and ensures that long-running backfills don't lock out other updates to the same table for the duration of the backfill.

Fixes #127

### DIFF
--- a/pkg/migrations/backfill.go
+++ b/pkg/migrations/backfill.go
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package migrations
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/schema"
+)
+
+// Backfill updates all rows in the given table, in batches, using the
+// following algorithm:
+// 1. Get the primary key column for the table.
+// 2. Get the first batch of rows from the table, ordered by the primary key.
+// 3. Update each row in the batch, setting the value of the primary key column to itself.
+// 4. Repeat steps 2 and 3 until no more rows are returned.
+func backfill(ctx context.Context, conn *sql.DB, table *schema.Table) error {
+	// Get the primary key column for the table
+	pks := table.GetPrimaryKey()
+	if len(pks) != 1 {
+		return errors.New("table must have a single primary key column")
+	}
+	pk := pks[0]
+
+	// Create a batcher for the table.
+	b := batcher{
+		table:     table,
+		pkColumn:  pk,
+		lastPK:    nil,
+		batchSize: 1000,
+	}
+
+	// Update each batch of rows
+	for {
+		if err := b.updateBatch(ctx, conn); err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				break
+			}
+			return err
+		}
+	}
+
+	return nil
+}
+
+type batcher struct {
+	table     *schema.Table
+	pkColumn  *schema.Column
+	lastPK    interface{}
+	batchSize int
+}
+
+// updateBatch updates the next batch of rows in the table.
+func (b *batcher) updateBatch(ctx context.Context, conn *sql.DB) error {
+	// Start the transaction for this batch
+	tx, err := conn.BeginTx(ctx, nil)
+	if err != nil {
+		return err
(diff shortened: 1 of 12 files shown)


pkg/migrations/backfill/backfill.go
```diff
@@ -0,0 +1,46 @@
...
+// following algorithm:
+// 1. Get the primary key column for the table.
+// 2. Get the first batch of rows from the table, ordered by the primary key.
+// 3. Update each row in the batch, setting the value of the primary key column to itself.
+// 4. Repeat steps 2 and 3 until no more rows are returned.
+func BackFill(ctx context.Context, conn *sql.DB, table *schema.Table) error {
+	// Get the primary key column for the table
+	pks := table.GetPrimaryKey()
+	if len(pks) != 1 {
+		return errors.New("table must have a single primary key column")
```
exekias, line 23: is this a hard requirement? multi column primary keys are not that strainge

pkg/migrations/backfill/batcher.go
```diff
@@ -0,0 +1,63 @@
...
+// buildQuery builds the query used to update the next batch of rows.
+func (b *batcher) buildQuery() string {
+	whereClause := ""
+	if b.lastPK != nil {
+		whereClause = fmt.Sprintf("WHERE %s > %v", pq.QuoteIdentifier(b.pkColumn.Name), b.lastPK)
+	}
+
+	return fmt.Sprintf(`
+    WITH batch AS (
+      SELECT %[1]s FROM %[2]s %[4]s ORDER BY id LIMIT %[3]d FOR UPDATE
```
exekias, line 53: is `ORDER BY id` correct? Is it guaranteed that an id column can always be used?

pkg/migrations/backfill/batcher.go
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package backfill
```
exekias, line 3: it would be good to have some test for this package

pkg/migrations/backfill/batcher.go
andrew-farries: Thanks; that should have been parameterized. I've rebased to fix it.

pkg/migrations/backfill/backfill.go
andrew-farries: I think the same approach could work with multi-column PKs, but creating multi-column PKs is not currently supported by `pgroll`. 

I'd rather leave it with this limitation in place and then test the multi-column PK behaviour when such tables are supported by `pgroll`. At the moment it's hard to write a test for as we can't set up such a table with `pgroll` migrations.

pkg/migrations/backfill/batcher.go
andrew-farries: The backfill behaviour is already tested by all the operations that perform backfills on start (setting `NOT NULL`, change type...).

Unless you think it's worth adding some larger performance tests for how long it takes against large table, I'd rather test the user-visibile behaviours that use backfill, rather than adding tests for backfill in isolation.

pkg/migrations/backfill/backfill.go
exekias: that sounds good for now, although pgroll can work against existing schemas. I believe at the very least we should detect this situation in validate so we don't attempt to start a migration that will result on this error

andrew-farries:
I've also moved the backfill code back into the `migrations` package as it's going to cause import cycles when we start to add progress updates for the `Start` operation.

pkg/migrations/backfill/backfill.go
andrew-farries: I've added extra validation checks for this case.

PR #146, opened 2023-09-27
Add docs for create table operation

andrew-farries:
Add a documentation section for the **create table** operation.

This is the first operation to be documented. If we're happy with this structure the other operations can be documented in a similar style.

[Direct link](https://github.com/xataio/pgroll/blob/doc//create-table-operation/docs/README.md#create-table)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -458,14 +458,14 @@ We've seen:
 
 `pgroll` migrations are specified as JSON files. All migrations follow the same basic structure:
 
-```go
-type Migration struct {
-    Name       string      `json:"name"`
-    Operations []Operation `json:"operations"`
+```json
+{
+  "name": "0x_migration_name",
+  "operations": [...]
 }
 ```
 
-In addition to this documentation the [examples](../examples) directory contains examples of each kind of operation.
+See the [examples](../examples) directory for examples of each kind of operation.
 
 `pgroll` supports the following migration operations:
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -480,6 +491,64 @@ We've seen:
...
+
+**create table** migrations have this structure:
+
+```go
+type OpCreateTable struct {
+	Name    string   `json:"name"`
+	Columns []Column `json:"columns"`
+}
+
+```
```
exekias, line 505: I would rather go for JSON based docs rather than Go. Many users won't know Go and that should be fine?

andrew-farries: * d4261d843522b27aa5ff6270108b8419b8a3d7f2 uses an ad-hoc JSON-style description of the operation.
* e405a8075b72773922b580506d661c1eda91bcaa uses Typescript descriptions.

Personally I prefer the latter, but happy to go with the former. Using [JSON Schema](https://json-schema.org/) would be another option but it seems excessively verbose to be useful as an informal description; it feels more geared towards schema verification.

exekias: I like the JSON-style more because it's language agnostic, users don't really need to know any programming language, but they need to understand JSON. Approving the PR so we don't block docs, I would go for https://github.com/xataio/pgroll/commit/d4261d843522b27aa5ff6270108b8419b8a3d7f2 and iterate from there

exekias (approved):

PR #150, opened 2023-09-29
Change `SELECT ... FOR UPDATE` to `NO KEY UPDATE` in backfill

andrew-farries:
Use a weaker lock in order to not block inserts into other tables that reference a locked row.

The rows locked for update by the `SELECT` will not have their primary keys modified so the weaker lock is sufficient.


exekias (approved):

exekias:
Awesome, also TIL :)

Issue #65, opened 2023-08-25
Take explicit locks when doing changes


exekias:
We are letting Postgres to take the proper locks when doing changes to schema objects (ie tables). This is fine in most cases but it can lead to database locking because of other queries happening during the migration.

To avoid this we should create and use helpers to get explicit locks, leveraging [lock timeout](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-LOCK-TIMEOUT) to ensure we never block access for too long.

some more info: https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#f0d8

exekias:
Closing as the PR fixing this has been merged

PR #155, opened 2023-10-02
Set session level `lock_timeout` for pgroll connection

andrew-farries:
Set the session `lock_timeout` for the connection used by `pgroll` to run DDL operations.

The need to set `lock_timeout` for DDL operations to achieve zero downtime migrations is well documented [[1](https://xata.io/blog/postgres-schema-changes-pita#locking-gotchas)], [[2](https://medium.com/paypal-tech/postgresql-at-scale-database-schema-changes-without-downtime-20d3749ed680#00dc)]. 

When a DDL transaction attempts to take an `ACCESS EXCLUSIVE` lock on a table while another transaction holds a conflicting lock (such as a long-running `SELECT` query's `ACCESS SHARE` lock),  other queries will queue up behind the DDL transaction until it is able to obtain the exclusive lock thus locking out reads from the table and causing downtime. By setting a `lock_timeout` the DDL transaction will abort if it is unable to acquire the lock before the timeout, freeing any other queries that were queued behind it.

The approach taken here is to set the `lock_timeout` at the session level. This assumes that `pgroll` has a direct connection to the Postgres instance rather than going through a connection pooler like PGBouncer. PGBouncer's statement or transaction mode may swap out the connection between transactions/statements meaning that the lock timeout may not be applied correctly. See [[3](https://jpcamara.com/2023/04/12/pgbouncer-is-useful.html)] for a more detailed discussion of PGBouncer's behaviour here.

The new `--lock-timeout` flag (and `PGROLL_LOCK_TIMEOUT` env var) is used to control the timeout duration.

Fixes https://github.com/xataio/pgroll/issues/65

### DIFF
--- a/cmd/flags/flags.go
+++ b/cmd/flags/flags.go
@@ -2,7 +2,9 @@
 
 package flags
 
-import "github.com/spf13/viper"
+import (
+	"github.com/spf13/viper"
+)
 
 func PostgresURL() string {
 	return viper.GetString("PG_URL")
@@ -14,4 +16,8 @@ func Schema() string {
 
 func StateSchema() string {
 	return viper.GetString("STATE_SCHEMA")
+}
+
+func LockTimeout() int {
+	return viper.GetInt("LOCK_TIMEOUT")
 }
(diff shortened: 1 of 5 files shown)


exekias:
LGTM! also needs the docs update

andrew-farries:
The `--lock-timeout` flag (and env var) is documented in the new `CLI reference` docs section: https://github.com/xataio/pgroll/pull/156

exekias:
sorry I merged this one by mistake, still ✅ so I think it should be fine. I'm thinking we should get the other 2 in and perhaps cut a new release?

PR #157, opened 2023-10-02
Add missing concepts to the docs

exekias:
This was the last missing peace to complete docs, hence I removed the "in progress" warning. Of course, we will keep improving them and addressing feedback, but I believe we can consider them complete.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -9,13 +9,31 @@
 * [Command line reference](#command-line-reference)
 * [Operations reference](#operations-reference)
 
-:warning: This section is under development.
-
 ## Concepts
+
+`pgroll` introduces a few concepts that are important to understand before using the tool.
 
 ### Migration workflow
 
+`pgroll` migrations are applied in two steps, following an [expand/contract pattern](https://openpracticelibrary.com/practice/expand-and-contract-pattern/).
+
+![migration flow](img/migration-flow@2x.png)
+
+During the migration start phase, `pgroll` will perform all additive changes to the database schema. This includes: creating new tables, adding new columns, and creating new indexes. In the cases were a required change is not backwards compatible, `pgroll` will take the necessary steps to ensure that the current schema is still valid. For example, if a new column is added to a table with a `NOT NULL` constraint, `pgroll` will backfill the new column with a default value.
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -9,14 +9,32 @@
...
 
+`pgroll` introduces a few concepts that are important to understand before using the tool.
+
 ### Migration workflow
 
+`pgroll` migrations are applied in two steps, following an [expand/contract pattern](https://openpracticelibrary.com/practice/expand-and-contract-pattern/).
+
+![migration flow](img/migration-flow@2x.png)
+
+During the migration start phase, `pgroll` will perform all additive changes to the database schema. This includes: creating new tables, adding new columns, and creating new indexes. In the cases were a required change is not backwards compatible, `pgroll` will take the necessary steps to ensure that the current schema is still valid. For example, if a new column is added to a table with a `NOT NULL` constraint, `pgroll` will backfill the new column with a default value.
```
andrew-farries, line 22:
```suggestion
During the migration start phase, `pgroll` will perform only additive changes to the database schema. This includes: creating new tables, adding new columns, and creating new indexes. In the cases were a required change is not backwards compatible, `pgroll` will take the necessary steps to ensure that the current schema is still valid. For example, if a new column is added to a table with a `NOT NULL` constraint, `pgroll` will backfill the new column with a default value.
```

docs/README.md
```diff
@@ -9,14 +9,32 @@
...
+
 ### Multiple schema versions
 
+`pgroll` maintains multiple versions of the database schema side-by-side. This is achieved by creating a new Postgres schema for each migration that is applied to the database. The schema will contain views on the underlying tables. These views are used to expose different tables or columns to client applications depending on which version of the schema they are configured to use.
+
+For instance, a rename column migration will create a new schema containing a view on the underlying table with the new column name. This allows for the new version of the schema to become available without breaking existing client applications that are still using the old name. On the migration complete phase, the old schema is dropped and the actual column is renamed (views are updated to point to the new column name automatically).
```
andrew-farries, line 32:
```suggestion
For instance, a rename column migration will create a new schema containing a view on the underlying table with the new column name. This allows for the new version of the schema to become available without breaking existing client applications that are still using the old name. In the migration complete phase, the old schema is dropped and the actual column is renamed (views are updated to point to the new column name automatically).
```

andrew-farries (approved):
This is really good 👍

Issue #179, opened 2023-10-06
Use `security_invoker` views for temporary views created in the new schema


ben-pr-p:
First, thanks for introducing this library - it's great to have, and I'm planning on contributing some extra tooling around it, specifically an integration with Kysely.

I think there's a bug (I haven't verified it, but just believe it's there from reading the code), that the views on the new schema that point to the tables in the old schema are created using the default security setting, which is similar to security definer.

This means that during the transition phase, row level security on the underlying tables will be bypassed by the views created by pgroll.

In order to make sure that accessing the old schema and new schema are the same, `create view view_name with (security_invoker = true) as ....` should be used.

See this article on the PG 15 release for more info: https://www.depesz.com/2022/03/22/waiting-for-postgresql-15-add-support-for-security-invoker-views/

I think that pgroll should also have some documented limitation on this potential "gotcha" in versions before Postgres 15, and potentially even error if there's no way around it (if there is an RLS policy on a table *and* the version is 14, which I see is supported).

Thanks again for all your work, super excited for what it means for the ecosystem!

exekias:
Thank you for bringing this up! I agree we should document better this and other limitations, happy to take a first stab at that.

Adding `security_invoker` to pgroll views makes sense to me, and should be not too complicated. We will need to retrieve the Postgres version first, as I believe this would be the first conditional clause we add based on it.

Also, really looking forward to knowing more about your plans for Kysely integration, this is great to hear! It seems there is enough interest on integrating ORMs & query builders across the board, and it’s something we want to look into: https://github.com/xataio/pgroll/issues/166

andrew-farries:
@ben-pr-p FYI we have a PR up that addresses the issue you raised with `security_invoker` views: https://github.com/xataio/pgroll/pull/189.

PR #189, opened 2023-10-17
Create versioned views `WITH (security_invoker = true)`

andrew-farries:
Ensure that views (in Postgres 15 and 16) are created with `security_invoker = true`.

This ensures that any row-level security on the underlying table is applied according to the permissions of the invoker of the view rather than its owner.

Postgres 14 does not support the creation of views with `security_invoker = true`.

See:
* https://www.depesz.com/2022/03/22/waiting-for-postgresql-15-add-support-for-security-invoker-views/
* https://www.cybertec-postgresql.com/en/view-permissions-and-row-level-security-in-postgresql/

Closes https://github.com/xataio/pgroll/issues/179

### DIFF
--- a/pkg/roll/execute.go
+++ b/pkg/roll/execute.go
@@ -147,16 +147,27 @@ func (m *Roll) Rollback(ctx context.Context) error {
 }
 
 // create view creates a view for the new version of the schema
-func (m *Roll) createView(ctx context.Context, version string, name string, table schema.Table) error {
+func (m *Roll) createView(ctx context.Context, version, name string, table schema.Table) error {
 	columns := make([]string, 0, len(table.Columns))
 	for k, v := range table.Columns {
 		columns = append(columns, fmt.Sprintf("%s AS %s", pq.QuoteIdentifier(v.Name), pq.QuoteIdentifier(k)))
 	}
 
+	// Create view with security_invoker option for PG 15+
+	//
+	// This ensures that any row level security permissions on the underlying
+	// table are respected. `security_invoker` views are not supported in PG 14
+	// and below.
+	withOptions := ""
+	if m.PGVersion() >= PGVersion15 {
+		withOptions = "WITH (security_invoker = true)"
+	}
+
 	_, err := m.pgConn.ExecContext(ctx,
-		fmt.Sprintf("CREATE OR REPLACE VIEW %s.%s AS SELECT %s FROM %s",
+		fmt.Sprintf("CREATE OR REPLACE VIEW %s.%s %s AS SELECT %s FROM %s",
 			pq.QuoteIdentifier(VersionedSchemaName(m.schema, version)),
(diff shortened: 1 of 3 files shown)


pkg/roll/roll.go
```diff
@@ -44,10 +45,23 @@ func New(ctx context.Context, pgURL, schema string, lockTimeoutMs int, state *st
...
+	}
+
 	return &Roll{
-		pgConn: conn,
-		schema: schema,
-		state:  state,
+		pgConn:      conn,
+		schema:      schema,
+		state:       state,
+		viewCreator: viewCreator,
```
exekias, line 64: not against this but I guess that we will do conditional statements based on the version in many places. Perhaps it would be good to store the version as a field here and read it from the createViews function to use one creation mode or another

.github/workflows/build.yml
```diff
@@ -15,11 +15,18 @@ jobs:
...
-    - name: Run tests
+    - name: Run all tests
+      if: matrix.pgVersion != '14.8'
       run: go test ./...
       env:
         POSTGRES_VERSION: ${{ matrix.pgVersion }}
 
+    - name: Run pg14 tests
+      if: matrix.pgVersion == '14.8'
+      run: go test ./... --skip 'TestViewsAreCreatedWithSecurityInvokerTrue'
```
exekias, line 26: Could we skip the test from the test case instead? calling `t.Skip` or similar based on the postgres version

andrew-farries: I've rebased to skip the test as you suggest rather than skipping in github actions.

pkg/roll/execute.go
```diff
@@ -147,16 +147,22 @@ func (m *Roll) Rollback(ctx context.Context) error {
...
 
 // create view creates a view for the new version of the schema
-func (m *Roll) createView(ctx context.Context, version string, name string, table schema.Table) error {
+func (m *Roll) createView(ctx context.Context, version, name string, table schema.Table) error {
 	columns := make([]string, 0, len(table.Columns))
 	for k, v := range table.Columns {
 		columns = append(columns, fmt.Sprintf("%s AS %s", pq.QuoteIdentifier(v.Name), pq.QuoteIdentifier(k)))
 	}
 
+	withOptions := ""
```
exekias, line 156: Could you add a comment here to explain security_invoker? 🙏

exekias (approved):
LGTM

My impression is that as we make progress (and new Postgres releases appear), we will do this more and more. Over the time more features will appear making zero downtime changes easier, or that I would hope

Having that into account, this approach looks good to me. As of today any option would work for me, my only worry is that if we solve this through dependency injection the `Roll` struct will only grow on logic that could be more local

pkg/roll/execute.go
andrew-farries: done

PR #191, opened 2023-10-18
Add `SECURITY DEFINER` to internal functions

andrew-farries:
We want to work on the security model for the `pgroll` schema, allowing admins to remove access while maintaining functionality intact. Because of this, some functions must still be marked as executable for any user doing migrations directly with postgres.

This PR adds `SECURITY DEFINER` to these functions

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -51,6 +51,7 @@ CREATE OR REPLACE FUNCTION %[1]s.is_active_migration_period(schemaname NAME) RET
 
 -- Get the latest version name (this is the one with child migrations)
 CREATE OR REPLACE FUNCTION %[1]s.latest_version(schemaname NAME) RETURNS text
+SECURITY DEFINER
 AS $$ 
   SELECT p.name FROM %[1]s.migrations p 
   WHERE NOT EXISTS (
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -172,7 +172,7 @@ BEGIN
 
 	ELSIF tg_event = 'ddl_command_end' THEN
 		-- Guess the schema from ddl commands, ignore migrations that touch several schemas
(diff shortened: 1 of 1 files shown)


pkg/state/state.go
```diff
@@ -51,6 +51,7 @@ CREATE OR REPLACE FUNCTION %[1]s.is_active_migration_period(schemaname NAME) RET
 
 -- Get the latest version name (this is the one with child migrations)
 CREATE OR REPLACE FUNCTION %[1]s.latest_version(schemaname NAME) RETURNS text
+SECURITY DEFINER
```
jankatins, line 54: This probably needs a search path and probably should also get some explicit schemas for all called functions, otherwise I can imagine that this can be used to gain the rights of the user used for migrations.

exekias: Thanks for raising this! I'm adding `SET search_path` and reviewing calls

exekias:
I took over this PR and updated the description, let me know what you think!

pkg/state/state.go
```diff
@@ -192,7 +196,7 @@ BEGIN
 	VALUES (
 		schemaname,
 		format('sql_%%s', substr(md5(random()::text), 0, 15)),
```
jankatins, line 198: While I currently think adding a search path is sufficient, I find it good from to use schema qualified function calls all over,  e.g. also `pg_catalog.format` and `pg_catalog.count(DISTINCT schema_name)`.

exekias:
Thank you @jankatins for your input! I believe I've addressed all comments

kvch (approved):

jankatins:
There are probably a few more which could be schema qualified (e.g. stuff like `>` is also overwriteable by a user which can manipulate the search path, as well as `md5()` and `random()`.

If you want to have some fun with testing/linting: https://github.com/Aiven-Open/pghostile and https://github.com/timescale/pgspot

PR #197, opened 2023-11-02
Fix previous version detection in the presence of inferred DDL migrations

### LINKED ISSUE #196: Inferred DDL migrations break previous version detection
When one or more inferred DDL migrations are captured after a `pgroll` migration has completed, the next `pgroll` migration to complete will fail to remove the version schema for the previous `pgroll` migration.

To reproduce:

* Run`pgroll start examples/01_create_tables.json --complete `
* Now run some DDL statements against the database, for example:
```sql
create table foo(a int);
create table bar(a int);
```
* Run `pgroll start examples/02_create_another_table.json --complete`

The `public_01_create_tables` schema should have been removed when the second migration was completed.

andrew-farries:
Update the definition of the `previous_version` function to use a recursive CTE to find the first non-'inferred' parent of the current version.

This requires adding a new column to the `pgroll.migrations` table to record whether a migration is a pgroll migration or an inferred DDL change made outside of pgroll.

Together, this means that the previous version schema is removed correctly when there have been inferred DDL changes between the current and previous pgroll migration.

Fixes #196

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -41,6 +41,13 @@ CREATE UNIQUE INDEX IF NOT EXISTS only_first_migration_without_parent ON %[1]s.m
 -- History is linear
 CREATE UNIQUE INDEX IF NOT EXISTS history_is_linear ON %[1]s.migrations (schema, parent);
 
+-- Add a column to tell whether the row represents an auto-detected DDL capture or a pgroll migration
+ALTER TABLE %[1]s.migrations ADD COLUMN IF NOT EXISTS migration_type
+  VARCHAR(32)
+  DEFAULT 'pgroll'
+  CONSTRAINT migration_type_check CHECK (migration_type IN ('pgroll', 'inferred')
+);
+
 -- Helper functions
 
 -- Are we in the middle of a migration?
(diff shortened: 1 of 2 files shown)


pkg/state/state.go
```diff
@@ -41,6 +41,13 @@ CREATE UNIQUE INDEX IF NOT EXISTS only_first_migration_without_parent ON %[1]s.m
 -- History is linear
 CREATE UNIQUE INDEX IF NOT EXISTS history_is_linear ON %[1]s.migrations (schema, parent);
 
+-- Add a column to tell whether the row represents an auto-detected DDL capture or a pgroll migration
+ALTER TABLE %[1]s.migrations ADD COLUMN IF NOT EXISTS migration_type
```
exekias, line 45: wondering if we should add it to the create table statement too, so we get a full representation in the `CREATE TABLE` statement + first time users don't go through 2 table changes

pkg/state/state.go
```diff
@@ -41,6 +41,13 @@ CREATE UNIQUE INDEX IF NOT EXISTS only_first_migration_without_parent ON %[1]s.m
...
+  VARCHAR(32)
```
exekias, line 46: would it make sense to use a boolean? do you plan for other uses here?

andrew-farries: I don't forsee other migration types, but I'm not sure if switching to a boolean here and losing the possibility of adding more is worth it.

pkg/state/state.go
andrew-farries: We could add it there. I did it this way because then the column definition is only in one place and there is no chance of existing users ending up with slightly different definition than than new users.

I also thought of this as more of a 'migration' being applied to the table.

exekias (approved):

PR #198, opened 2023-11-03
Fix primary keys retrieval query

exekias:
The previous query had some issues when used under clusters with many schema objects, this one improves query performance


sorintm (approved):
LGTM

andrew-farries (approved):
Looks good to me.

Primary key lookup is checked by the testcases that test backfills, so if the tests pass and it's faster, then ✅

PR #205, opened 2023-11-22
Move logic for the `status` command into the `state` and `migrations` packages

andrew-farries:
Move the logic for the `pgroll status` command out of the CLI and into the `migrations` and `state` package and add tests for it.

This makes it possible to consume migration status information from packages using `pgroll` as a module.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -349,13 +349,13 @@ func (s *State) Status(ctx context.Context, schema string) (*Status, error) {
 		return nil, err
 	}
 
-	var status string
+	var status MigrationStatus
 	if *latestVersion == "" {
-		status = "No migrations"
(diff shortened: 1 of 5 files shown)


pkg/state/state.go
```diff
@@ -334,6 +334,37 @@ func (s *State) PreviousVersion(ctx context.Context, schema string) (*string, er
...
+		return nil, err
+	}
+	if latestVersion == nil {
+		latestVersion = new(string)
+	}
+
+	isActive, err := s.IsActiveMigrationPeriod(ctx, schema)
+	if err != nil {
+		return nil, err
+	}
```
exekias, line 350: not needed for this PR but we may want to retrieve these two with a single query in the future to guarantee consistency

exekias (approved):

PR #210, opened 2023-11-27
Add JSON schema struct generation

SferaDev:
All structs matched well except for `PrimaryKey` which we mapped differently, I've refactored it to `Pk` as it doesn't seem to support different names for go and JSON


andrew-farries (commented):
Looks good 👍 

* I think we should also add a `Makefile` with a rule to regenerate these types.
* The docs could be updated to include a reference to this JSON schema in the [operations reference section](https://github.com/xataio/pgroll/blob/main/docs/README.md#operations-reference)

schema.json
```diff
@@ -0,0 +1,388 @@
+{
+  "$ref": "#/definitions/PgRollMigration",
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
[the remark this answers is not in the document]
SferaDev: I will update

SferaDev:
@andrew-farries Updated!

- When doing `go get` for the dependency, it updated other versions in the mod/sum files.
- Added a `Makefile` with a `generate` step
- Fixed the license check in the `Makefile` step
- Updated all objects to include descriptions

Probably in the future we can add linter/formatter to the `Makefile`, but I'd leave that as a follow-up.

Makefile
```diff
@@ -0,0 +1,8 @@
+.PHONY: generate
+generate:
+	# Generate the types from the JSON schema
+	go-jsonschema --only-models -p migrations --tags json schema.json > pkg/migrations/types.go
```
andrew-farries, line 4: We could run `go-jsonschema` via docker to minimize dependencies and ensure a consistent version:

```suggestion
	docker run -v $$PWD/schema.json:/mnt/schema.json omissis/go-jsonschema:0.14.1 --only-models -p migrations --tags json /mnt/schema.json > pkg/migrations/types.go
```

schema.json
```diff
@@ -0,0 +1,404 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "allOf": [{ "$ref": "#/$defs/PgRollMigration" }],
```
andrew-farries, line 3: What's the purpose of `allOf` here? The same types are still generated successfully if I remove this line.

schema.json
```diff
@@ -0,0 +1,404 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
```
andrew-farries, line 2: should the schema have an `$id`? It seems standard on other JSON schema documents.

andrew-farries: Also how about a `title` and a `description`?

andrew-farries (commented):
This looks good, just a few more suggestions.

andrew-farries:
> Update docs and use the JSON schema to generate API reference

Do you have an idea in mind for how to generate documentation from the JSON schema?

SferaDev:
> > Update docs and use the JSON schema to generate API reference
> 
> Do you have an idea in mind for how to generate documentation from the JSON schema?

There are different tools to read JSON schemas and generate codes, one that comes to mind is the one that Cloudflare internally uses for their API documentation: https://github.com/cloudflare/json-schema-tools/tree/master/workspaces/doca

Although I was thinking more of doing something a little bit more tailored for xata.io/docs. I think we will likely have 2 documentation sites: pgroll open source repo (and maybe even website in a future), and xata.io website for documenting the schema migrations in-product. It would be great if we make both share a single source of truth like the JSON Schema.

schema.json
andrew-farries: The `PgRollMigration` type actually duplicates a type we define here:

https://github.com/xataio/pgroll/blob/main/pkg/migrations/migrations.go#L46-L53

So it could probably be removed entirely.

SferaDev: I missed those. What about I create: `Migration`, `Operation` and `Operations` and keep the `allOf` with `$ref` top level?

I can get it done tomorrow so it's ready for a final review.

SferaDev: After checking today, it might not be possible to generate these and re-use them.

1. The `Operation` in pgroll it's not just an struct, it's an interface with methods.
2. The library hasn't added support for `anyOf` yet, they have a [PR open](https://github.com/omissis/go-jsonschema/pull/99) but stale

I think it's fine as-is leaving the top level `allOf: { $ref }` to reference the definition of `PgRollMigration` that doesn't clash with the internal `Migration` one.

I've subscribed to that PR, once they merge support for `anyOf` maybe we can accomodate both and fully re-use `Migration`, `Operations` and `Operation`.

WDYT?

andrew-farries: Thats fine from my POV. The main benefit we get from this is being able to generate the types for the different operations. 

Is this PR good to go from your side now?

SferaDev: Yes, can be merged when you're good with it

andrew-farries (approved):

PR #212, opened 2023-12-05
docs: Clarify rules for quoting string literals

andrew-farries:
Update the documentation to clarify how to specify default values for the add column and create table operations.


exekias (approved):
perhaps it's worth adding an example? I believe the most common use case is to add a column without a default but with a fixed backfill value (up migration)

PR #224, opened 2024-01-10
Add support for creating tables and columns with comments

andrew-farries:
Update the **create table** and **add column** operations so that they support adding [Postgres comments](https://www.postgresql.org/docs/current/sql-comment.html):

* **create table**: comments can be added to the table itself and to each of its columns.
* **add column**: a comment can be added to the column.

A **create table** migration that includes a comment on the table itself and on one of its columns looks like this:

```json
{
  "name": "12_create_employees_table",
  "operations": [
    {
      "create_table": {
        "name": "employees",
        "comment": "This is a comment for the employees table",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "role",
            "type": "varchar(255)",
            "comment": "This is a comment for the role column"
          }
        ]
      }
    }
  ]
}
```

and an **add column** migration that includes a comment looks like this:

```json
{
  "name": "30_add_column_simple_up",
  "operations": [
    {
      "add_column": {
        "table": "people",
        "up": "'temporary-description'",
        "column": {
          "name": "description",
          "type": "varchar(255)",
          "nullable": false,
          "comment": "This is a comment for the description column"
        }
      }
    }
  ]
}
```

This allows new tables and columns to be created with comments.

Deletion and modification of comments should still be performed with a raw SQL migration. Until we see a use case that requires versioned modification/removal of comments, these operations are best performed directly on the base table with raw SQL.

### DIFF
--- a/pkg/migrations/op_common_test.go
+++ b/pkg/migrations/op_common_test.go
@@ -146,6 +146,20 @@ func ColumnMustHaveType(t *testing.T, db *sql.DB, schema, table, column, expecte
 	}
 }
 
+func ColumnMustHaveComment(t *testing.T, db *sql.DB, schema, table, column, expectedComment string) {
+	t.Helper()
+	if !columnHasComment(t, db, schema, table, column, expectedComment) {
+		t.Fatalf("Expected column %q to have comment %q", column, expectedComment)
+	}
+}
+
+func TableMustHaveComment(t *testing.T, db *sql.DB, schema, table, expectedComment string) {
+	t.Helper()
+	if !tableHasComment(t, db, schema, table, expectedComment) {
+		t.Fatalf("Expected table %q to have comment %q", table, expectedComment)
+	}
+}
+
 func TableMustHaveColumnCount(t *testing.T, db *sql.DB, schema, table string, n int) {
 	t.Helper()
 	if !tableMustHaveColumnCount(t, db, schema, table, n) {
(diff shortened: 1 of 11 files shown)


pkg/migrations/op_common_test.go
```diff
@@ -400,6 +414,40 @@ func columnHasType(t *testing.T, db *sql.DB, schema, table, column, expectedType
...
 
+func columnHasComment(t *testing.T, db *sql.DB, schema, table, column, expectedComment string) bool {
+	t.Helper()
+
+	var actualComment string
+	err := db.QueryRow(fmt.Sprintf(`
+    SELECT col_description(
+      %[1]s::regclass,
+      (SELECT attnum FROM pg_attribute WHERE attname=%[2]s and attrelid=%[1]s::regclass)
+    )`,
```
exekias, line 425: should we add this to the read_schema function?

andrew-farries: Looks like it's already there:

https://github.com/xataio/pgroll/blob/f1757f31a270c593e45b67ea79cd38af5c16cc21/pkg/state/state.go#L105

for table comments 

and 

https://github.com/xataio/pgroll/blob/f1757f31a270c593e45b67ea79cd38af5c16cc21/pkg/state/state.go#L129

for column comments.

Example output for one table:

```json
{
  "employees": {
    "oid": "16629",
    "name": "employees",
    "columns": {
      "id": {
        "name": "id",
        "type": "integer",
        "unique": true,
        "comment": null,
        "default": "nextval('_pgroll_new_employees_id_seq'::regclass)",
        "nullable": false
      },
      "job_title": {
        "name": "job_title",
        "type": "varchar(255)",
        "unique": false,
        "comment": "this is a comment for the role column",
        "default": null,
        "nullable": false
      }
    },
    "comment": "this is a comment for the employees table",
    "indexes": {
      "_pgroll_new_employees_pkey": {
        "name": "_pgroll_new_employees_pkey"
      }
    },
    "primarykey": [
      "id"
    ],
    "foreignkeys": null
  }
}
```

exekias (approved):

PR #233, opened 2024-01-15
Allow to disable version schemas

exekias:
There are situatiosn were creating schema verisons is not required. This internal flag will allow to disable creating & managing these schemas when needed.

Note: I opted to not expose this flag to the CLI as I believe it would make create more harm & confusion than not having the feature there.


andrew-farries (commented):
Code looks fine, apart from one suggestion around the tests.

In what situations do you imagine we will need the flag?

PR #245, opened 2024-01-18
Identify duplicated foreign key constraints by prefix

andrew-farries:
When duplicating a column for backfilling, ensure that any FK constraints on the column are duplicated with a special prefix to identify them as duplicated constraints. Naming the duplicated FK constraints in this way allows the rename process to ensure that it only attempts to rename duplicated FK constraints.


exekias (approved):
LGTM, how could a column end up with a fk constraint that doesn't come from duplication? I guess it's in the case you are actually adding it, which requires a duplicate?

andrew-farries:
> I guess it's in the case you are actually adding it, which requires a duplicate?

Yes, that's it. We need to distinguish the case where the FK on a temporary column is a duplicate because an FK was already on the column when we set the column `NOT NULL` for example, vs the case where the FK is on the temporary column because it's new on the column (ie put there by a 'set FK' operation).

Issue #227, opened 2024-01-12
Ensure faithful duplication of columns for backfilling


andrew-farries:
Various `pgroll` operations need to duplicate (and backfill) a column on migration start. The general pattern for these operations is:

On migration start:
* Create a duplicate of the affected column
* Backfill the duplicated column with values from the original.

On migration complete:
* Drop the original column
* Rename the duplicated column to the original column name

Duplication of columns currently loses a lot of information about the source column. These means that when such a migration completes, information (such as constraints and defaults) on the original column is lost. Currently only the column's type is preserved.

The operations that duplicate a column are:
- set `NOT NULL`.
- change type. 
-  drop a constraint.
-  add a `CHECK` constraint.
-  add a `FOREIGN KEY` constraint
-   add a `UNIQUE` constraint

The information that needs to be preserved when duplicating a column for backfilling is:

- column`DEFAULT`.
- foreign keys defined on the column. 
-   indexes defined on the column.
-  constraints defined on the column.
-   nullability

### Support matrix

For each property of a column, **p**,  and each operation type, **o**, we need to preserve property **p** when running operation **o**:

| **property**/**operation** | set `NOT NULL` |  drop `NOT NULL` | change type | drop constraint | add `CHECK` constraint | add `FK` constraint | add `UNIQUE` constraint |
|----------------------------|----------------|------------------|-------------|-----------------|------------------------|---------------------|-------------------------|
| column `DEFAULT`           | 🟢            | 🟢              | 🟢           | 🟢             | 🟢                    | 🟢                 | 🟢                     |
| FKs on column              | 🟢            | 🟢              | 🟢           | 🟢             | 🟢                    | 🟢                 | 🟢                     |
| indexes on column          | 🟢            | 🟢              | 🟢           | 🟢             | 🟢                    | 🟢                 | 🟢                     |
| check constraints on column| 🟢            | 🟢              | 🟢           | 🟢             | 🟢                    | 🟢                 | 🟢                     |
| unique constraints on column | 🟢          | 🟢              | 🟢           | 🟢             | 🟢                    | 🟢                 | 🟢                     |
| column nullability         | n/a           | n/a            | 🟢           | 🟢             | 🟢                    | 🟢                 | 🟢                     |
| column comments | 🟢             | 🟢               | 🟢            | 🟢              | 🟢                     | 🟢                  | 🟢                      |

kvch:
I assigned myself to the issue because I am adding the missing indices of a duplicated column.

[...]

PR #247, opened 2024-01-22
Record index uniqueness in `pgroll`'s internal schema representation

andrew-farries:
Add uniqueness information to each index recorded in `pgroll`'s internal schema representation.

With the extra information, the entry for indexes in the schema representation now looks like:

```json
"indexes": {                                                           
    "products_name_unique": {                                          
        "name": "products_name_unique",                                
        "unique": true,                                                                                                              
    },                                                                  
    "_pgroll_new_products_pkey": {                                     
        "name": "_pgroll_new_products_pkey",                           
        "unique": true,                                                                                                          
    }                                                                  
}
```

Having index uniqueness recorded in the schema representation will help with preserving uniqueness constraints on columns duplicated for backfilling (see https://github.com/xataio/pgroll/issues/227).

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -168,7 +168,8 @@ BEGIN
 				),
 				'indexes', (
 				  SELECT json_object_agg(pi.indexrelid::regclass, json_build_object(
-				    'name', pi.indexrelid::regclass
+				    'name', pi.indexrelid::regclass,
+				    'unique', pi.indisunique
 				  ))
 				  FROM pg_index pi 
 				  WHERE pi.indrelid = t.oid::regclass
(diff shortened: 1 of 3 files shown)


exekias (approved):

pkg/state/state.go
```diff
@@ -168,7 +168,8 @@ BEGIN
 				),
 				'indexes', (
 				  SELECT json_object_agg(pi.indexrelid::regclass, json_build_object(
-				    'name', pi.indexrelid::regclass
+				    'name', pi.indexrelid::regclass,
+				    'unique', pi.indisunique
```
exekias, line 172: are we missing the columns on what the index is created here?

andrew-farries: yes, that's coming in another PR.

PR #258, opened 2024-01-31
Add drop `NOT NULL` operation to remove `NOT NULL` from a column

### LINKED ISSUE #223: Support removal of not null constraint
When a column is added with a `notNull` constraint, removing it throws the following error: `Failed to start migration: migration is invalid: removing NOT NULL constraints is not supported `

Steps to reproduce:
1. Create tables:
```
{
    "name": "01_create_tables",
    "operations": [
      {
        "create_table": {
          "name": "bills",
          "columns": [
            {
              "name": "id",
              "type": "integer",
              "pk": true
            },
            {
              "name": "str",
              "type": "text",
              "nullable": true
            }
          ]
        }
      }
    ]
  }
```

2. Try and remove not null constraint
```
{
  "name": "17_set_nullable",
  "operations": [
    {
      "alter_column": {
        "table": "bills",
        "column": "str",
        "nullable": false, 
        "up": "''",
        "down": "''"
      }
    }
  ]
}
```

Maybe the `drop_constraint` syntax could be used instead?

Internal discussion: https://xata-hq.slack.com/archives/C05TR0HPQ9Y/p1702291736231379

andrew-farries:
Add a new sub-operation to the 'alter column' operation to drop `NOT NULL` constraints from columns.

Currently, it is only possible to **set** `NOT NULL` constraints on columns but not remove them. The syntax for doing this is:

```json
{
  "name": "16_set_nullable",
  "operations": [
    {
      "alter_column": {
        "table": "reviews",
        "column": "review",
        "nullable": false,
        "up": "(SELECT CASE WHEN review IS NULL THEN product || ' is good' ELSE review END)",
        "down": "review"
      }
    }
  ]
}
```

Setting `nullable: true` in the above migration would result in a validation error saying that removing `NOT NULL` constraints was not supported. This PR removes this restriction and allows `nullable: true`  to remove an existing `NOT NULL` constraint.

A migration to remove a `NOT NULL` constraint looks like:

```json
{
  "name": "31_unset_not_null",
  "operations": [
    {
      "alter_column": {
        "table": "posts",
        "column": "title",
        "nullable": true,
        "up": "title",
        "down": "(SELECT CASE WHEN title IS NULL THEN 'placeholder title' ELSE title END)"
      }
    }
  ]
}
```

The differences between adding and removing a `NOT NULL` constraint are:
* `nullable: true` vs `nullable: false`
* The roles of `up` and `down` SQL are reversed; `down` now needs to rewrite any `NULL`s to meet the `NOT NULL` constraint on the old column. `up` defaults to a simple copy of the data from the old column to the new.

Fixes #223

### DIFF
--- a/pkg/migrations/op_drop_not_null.go
+++ b/pkg/migrations/op_drop_not_null.go
@@ -1,0 +1,157 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package migrations
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/schema"
+)
+
+type OpDropNotNull struct {
+	Table  string `json:"table"`
+	Column string `json:"column"`
+	Up     string `json:"up"`
+	Down   string `json:"down"`
+}
+
+var _ Operation = (*OpDropNotNull)(nil)
+
+func (o *OpDropNotNull) Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) error {
+	table := s.GetTable(o.Table)
+	column := table.GetColumn(o.Column)
+
+	// Create a copy of the column on the underlying table.
+	d := NewColumnDuplicator(conn, table, column).WithoutNotNull()
+	if err := d.Duplicate(ctx); err != nil {
+		return fmt.Errorf("failed to duplicate column: %w", err)
+	}
+
+	// Add a trigger to copy values from the old column to the new, rewriting NULL values using the `up` SQL.
+	err := createTrigger(ctx, conn, triggerConfig{
+		Name:           TriggerName(o.Table, o.Column),
+		Direction:      TriggerDirectionUp,
+		Columns:        table.Columns,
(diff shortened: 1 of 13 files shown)


pkg/migrations/op_drop_not_null.go
```diff
@@ -0,0 +1,157 @@
...
+	table := s.GetTable(o.Table)
+	column := table.GetColumn(o.Column)
+
+	// Create a copy of the column on the underlying table.
+	d := NewColumnDuplicator(conn, table, column).WithoutNotNull()
+	if err := d.Duplicate(ctx); err != nil {
+		return fmt.Errorf("failed to duplicate column: %w", err)
+	}
+
+	// Add a trigger to copy values from the old column to the new, rewriting NULL values using the `up` SQL.
```
exekias, line 33: technically it's rewriting all values 😇 it will be NULL ones only if the up function does the typical `CASE`.

exekias (approved):

pkg/migrations/op_drop_not_null.go
andrew-farries: Thanks, I've updated the comment.

PR #259, opened 2024-01-31
Fix inferred migrations format

exekias:
Inferred migrations do not follow the same format as all other stored migrations. From the schema history point of view, all migrations in the list should apply correctly, including these.

This change ensures we store them following the same format, so replaying a migration history is possible.

### DIFF
--- a/pkg/state/state_test.go
+++ b/pkg/state/state_test.go
@@ -5,10 +5,13 @@ package state_test
 import (
 	"context"
 	"database/sql"
+	"encoding/json"
+	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"github.com/lib/pq"
 	"github.com/stretchr/testify/assert"
 	"github.com/xataio/pgroll/pkg/migrations"
 	"github.com/xataio/pgroll/pkg/schema"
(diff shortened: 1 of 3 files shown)


pkg/state/state_test.go
```diff
@@ -56,6 +59,70 @@ func TestSchemaOptionIsRespected(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		}
+
+		// init the state
+		if err := state.Init(ctx); err != nil {
+			t.Fatal(err)
+		}
```
andrew-farries, line 89: I think this state initalization should be moved inside the `WithStateAndConnectionToContainer` helper. The other usages of this helper all init the state outside too.

pkg/state/state_test.go
```diff
@@ -56,6 +59,70 @@ func TestSchemaOptionIsRespected(t *testing.T) {
...
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				if _, err := db.ExecContext(ctx, "DROP SCHEMA public CASCADE; CREATE SCHEMA public"); err != nil {
+					t.Fatal(err)
+				}
```
andrew-farries, line 95: Why is it necessary to drop and recreate the public schema?

pkg/state/state_test.go
```diff
@@ -56,6 +59,70 @@ func TestSchemaOptionIsRespected(t *testing.T) {
...
+
+				var gotMigration migrations.Migration
+				if err := json.Unmarshal(migrationStr, &gotMigration); err != nil {
+					t.Fatal(err)
+				}
+
+				// test there is a name for the migration, then remove it for the comparison
+				assert.True(t, len(gotMigration.Name) > 1)
```
andrew-farries, line 115: We could use a stricter assertion here; longer length and prefixed with 'sql_'

pkg/state/state_test.go
exekias: this is a table based test, this will avoid future tests colliding between then

andrew-farries: in that case I think it would be clearer if each testcase created a schema for itself. So the testcase definition becomes:

```go
struct {
  name          string
  sqlStmt       string
  schema        string
  wantMigration migrations.Migration
}
```

andrew-farries (commented):
One more comment about dropping and recreating the `public` schema.

andrew-farries (approved):

PR #261, opened 2024-02-01
Update `schema.json` to correctly describe the 'alter column' operation

### LINKED ISSUE #222: Generated `alter_column` input type for async migration endpoint is too strict
### What happened?

When calling the async migration endpoint, `type` and `name` arguments are [required](https://github.com/xataio/pgroll/blob/main/schema.json#L149) on the input. At run time, specifying both causes the migration to fail with: `migration is invalid: alter column operations require exactly one change, found 2.`

Here is where the check occurs:
https://github.com/xataio/pgroll/blob/e3faa35de95a1861ec88b4c05bd5df422177cfca/pkg/migrations/op_alter_column.go#L34-L37

### What did you expect to see?

I expected to see the generated types for the input argument here not require both `name` and `type` argument.

andrew-farries:
Update `schema.json` to correctly describe the 'alter column' operation.

The [alter column ](https://github.com/xataio/pgroll/tree/main/docs#alter-column) operation is used to perform different alter column operations according to which fields are present.

The rules are:
* `table` and `column` are always required.
* Exactly one of `name`, `type`, `check`, `unique`, `references`, `nullable` must be present.
* If `type`, `check`, `unique`, `references` or `nullable` is present, then `up` and `down` must also be present.
* If `name` is present then neither `up` or `down` is allowed.

This PR updates `schema.json` to capture these dependencies.

This change causes churn in the test code as some fields that previously `string` have become `*string`.

Fixes #222


schema.json
```diff
@@ -150,7 +150,28 @@
...
         }
       },
-      "required": ["column", "down", "name", "table", "type", "up"],
+      "required": ["table", "column"],
+      "oneOf": [
+        {
+          "required": ["name"],
+          "not": {
+            "required": ["up", "down"]
+          }
```
SferaDev, line 159: I think this is not needed

schema.json
```diff
@@ -150,7 +150,28 @@
...
+          "oneOf": [
+            {"required": ["check"]},
+            {"required": ["type"]},
+            {"required": ["nullable"]},
+            {"required": ["unique"]},
+            {"required": ["references"]}
+          ],
+          "not": {
+            "required": ["name"]
+          }
```
SferaDev, line 172: I think this is not needed

andrew-farries: Removing it allows this, which isn't valid:

```json
{
  "table": "my_table",
  "column": "my_column",
  "name": "foo",
  "up": "foo",
  "down": "foo",
}
```

`up` and `down` aren't valid with `name`.

schema.json
andrew-farries: If this is removed then the following invalid `json` validates:

```json
{
  "table": "my_table",
  "column": "my_column",
  "name": "foo",
  "up": "foo",
  "down": "foo",
}
```

`up` and `down` are invalid when `name` is specified.

SferaDev: But it shouldn't be `not.required`, the property should not be defined outside but inside the other part, right?

andrew-farries: I'm not sure what you mean here. Could you post what you think this `oneOf` section of the spec should look like?

SferaDev: Sure, give me 30 min to finish a couple of things and I'll make a suggestion comment

andrew-farries:
I think we should add some tests around schema validation for the 'alter column' operation here too.

andrew-farries:
I pushed another update to the schema - it fixes the problem where `name` could be specified alongside other operations like `check`, `nullable` etc.

andrew-farries:
I've added some tests for JSON schema validation in https://github.com/xataio/pgroll/pull/263

exekias (approved):

SferaDev (approved):
Spec also LGTM

PR #265, opened 2024-02-02
Update JSON schema to specify that not all `Column` fields are required

andrew-farries:
Update the JSON schema in `schema.json` to state that only `name` and `type` are required fields when specifying a column. 

Previously `required` was set to `["name", "nullable", "pk", "type", "unique"]` resulting in some valid `create_table` and `add_column` migrations failing to validate.

Add two new testcases to the JSON schema tests to cover 'create_table' and 'add_column' operations.


SferaDev:
Spec LGTM

exekias (approved):

PR #274, opened 2024-02-05
Update `state` command to avoid returning nulls

SferaDev:
When building the OpenAPI spec we noticed that there were a lot of empty dictionaries and arrays.
Updating the `state.go` query to default for `{}` and `[]`.


exekias:
I think this would be the way to go for tables & all other dicts 💪

exekias (approved):

Issue #64, opened 2023-08-25
Add guards for names max length


exekias:
names in postgres have a max length of 63 chars by default. We generate names in many places, like:

- Version schema names
- Temporary table/column names
- indices
- triggers
- etc

We should make sure we don't generate names longer than the expected max, as that could cause issues down the line.

https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

pucke-uninow:
I would like to work on this one.

PR #275, opened 2024-02-05
Max length for naming

0del:
TODO list:
- [x] Apply max length validation for temporary name
- [x] Operations validation
- [x] Schema name validation
- [x] triggers name validation

fixes: https://github.com/xataio/pgroll/issues/64

### DIFF
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -177,9 +177,9 @@ func (e InvalidReplicaIdentityError) Error() string {
 
 type InvalidNameLengthError struct {
 	Name string
-	Max      int
+	Max  int
 }
 
 func (e InvalidNameLengthError) Error() string {
-	return fmt.Sprintf("max length of %q is %d", e.Identity, e.Max)
+	return fmt.Sprintf("max length of %q is %d", e.Name, e.Max)
 }
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -157,13 +157,12 @@ func (e MultipleAlterColumnChangesError) Error() string {
 	return fmt.Sprintf("alter column operations require exactly one change, found %d", e.Changes)
 }
 
-type InvalidPrimaryKeyError struct {
-	Table  string
-	Fields int
+type BackfillNotPossibleError struct {
+	Table string
 }
 
-func (e InvalidPrimaryKeyError) Error() string {
-	return fmt.Sprintf("primary key on table %q must be defined on exactly one column, found %d", e.Table, e.Fields)
+func (e BackfillNotPossibleError) Error() string {
+	return fmt.Sprintf("a backfill is required but table %q doesn't have a single column primary key or a UNIQUE, NOT NULL column", e.Table)
 }
 
 type InvalidReplicaIdentityError struct {
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -176,7 +176,7 @@ func (e InvalidReplicaIdentityError) Error() string {
 }
 
 type InvalidNameLengthError struct {
-	Identity string
+	Name string
 	Max      int
 }
 
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -181,5 +181,5 @@ type InvalidNameLengthError struct {
 }
 
 func (e InvalidNameLengthError) Error() string {
-	return fmt.Sprintf("%d", e.Max)
+	return fmt.Sprintf("max length of %s is %d", e.Identity, e.Max)
 }
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -181,5 +181,5 @@ type InvalidNameLengthError struct {
 }
 
 func (e InvalidNameLengthError) Error() string {
-	return fmt.Sprintf("max length of %s is %d", e.Identity, e.Max)
+	return fmt.Sprintf("max length of %q is %d", e.Identity, e.Max)
 }
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -174,3 +174,12 @@ type InvalidReplicaIdentityError struct {
 func (e InvalidReplicaIdentityError) Error() string {
 	return fmt.Sprintf("replica identity on table %q must be one of 'NOTHING', 'DEFAULT', 'INDEX' or 'FULL', found %q", e.Table, e.Identity)
 }
+
+type InvalidNameLengthError struct {
+	Identity string
+	Max      int
+}
+
+func (e InvalidNameLengthError) Error() string {
+	return fmt.Sprintf("%d", e.Max)
+}
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -137,6 +137,15 @@ func (e ConstraintDoesNotExistError) Error() string {
 	return fmt.Sprintf("constraint %q on table %q does not exist", e.Constraint, e.Table)
 }
 
+type ConstraintAlreadyExistsError struct {
+	Table      string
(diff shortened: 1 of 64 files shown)


pkg/migrations/op_set_fk.go
```diff
@@ -149,6 +149,13 @@ func (o *OpSetForeignKey) Rollback(ctx context.Context, conn *sql.DB) error {
 }
 
 func (o *OpSetForeignKey) Validate(ctx context.Context, s *schema.Schema) error {
+	name := TriggerName(o.Table, TemporaryName(o.Column))
+	if len(name) > MaxNameLength {
+		return InvalidNameLengthError{
+			Identity: name,
+			Max:      MaxNameLength,
+		}
+	}
```
andrew-farries, line 158: I don't think it makes sense to check the names of the triggers. The trigger functions are internal to `pgroll` and their names won't mean anything to the users starting the migration.

I would remove all the checks for trigger name length and just check the names of the user-defined parts of migrations like column and constraint names.

SferaDev:
@andrew-farries Updated with main and applied the changes to make CI checks pass

pkg/migrations/errors.go
```diff
@@ -183,6 +183,15 @@ func (e InvalidReplicaIdentityError) Error() string {
 	return fmt.Sprintf("replica identity on table %q must be one of 'NOTHING', 'DEFAULT', 'INDEX' or 'FULL', found %q", e.Table, e.Identity)
 }
 
+type InvalidNameLengthError struct {
+	Name string
+	Max  int
+}
+
+func (e InvalidNameLengthError) Error() string {
+	return fmt.Sprintf("max length of %q is %d", e.Name, e.Max)
```
andrew-farries, line 192:
```suggestion
	return fmt.Sprintf("length of %q (%d) exceeds maximum length of %d", e.Name, len(e.Name), e.Max)
```

pkg/migrations/name.go
```diff
@@ -0,0 +1,17 @@
...
+func validateName(name string) error {
+	if len(name) > MaxNameLength {
+		return InvalidNameLengthError{Name: name, Max: MaxNameLength}
+	}
+	return nil
+}
+
+func ValidateName(name string) error {
+	return validateName(name)
+}
```
andrew-farries, line 17: Get rid of `validateName` and use the exported `ValidateName` everywhere.

pkg/migrations/op_change_type.go
```diff
@@ -138,5 +138,10 @@ func (o *OpChangeType) Validate(ctx context.Context, s *schema.Schema) error {
 	if o.Down == "" {
 		return FieldRequiredError{Name: "down"}
 	}
+
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
+
```
andrew-farries, line 145: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/migrations/op_drop_column.go
```diff
@@ -56,6 +56,9 @@ func (o *OpDropColumn) Rollback(ctx context.Context, conn *sql.DB) error {
 }
 
 func (o *OpDropColumn) Validate(ctx context.Context, s *schema.Schema) error {
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
```
andrew-farries, line 61: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/migrations/op_drop_constraint.go
```diff
@@ -122,6 +122,10 @@ func (o *OpDropConstraint) Rollback(ctx context.Context, conn *sql.DB) error {
 }
 
 func (o *OpDropConstraint) Validate(ctx context.Context, s *schema.Schema) error {
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
+
```
andrew-farries, line 128: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/migrations/op_drop_not_null.go
```diff
@@ -130,6 +130,10 @@ func (o *OpDropNotNull) Rollback(ctx context.Context, conn *sql.DB) error {
 }
 
 func (o *OpDropNotNull) Validate(ctx context.Context, s *schema.Schema) error {
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
+
```
andrew-farries, line 136: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/migrations/op_set_check.go
```diff
@@ -144,6 +144,10 @@ func (o *OpSetCheckConstraint) Rollback(ctx context.Context, conn *sql.DB) error
 }
 
 func (o *OpSetCheckConstraint) Validate(ctx context.Context, s *schema.Schema) error {
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
+
```
andrew-farries, line 150: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/migrations/op_set_fk.go
```diff
@@ -145,6 +145,9 @@ func (o *OpSetForeignKey) Rollback(ctx context.Context, conn *sql.DB) error {
 }
 
 func (o *OpSetForeignKey) Validate(ctx context.Context, s *schema.Schema) error {
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
```
andrew-farries, line 150: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/migrations/op_set_notnull.go
```diff
@@ -162,6 +162,10 @@ func (o *OpSetNotNull) Rollback(ctx context.Context, conn *sql.DB) error {
 }
 
 func (o *OpSetNotNull) Validate(ctx context.Context, s *schema.Schema) error {
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
+
```
andrew-farries, line 168: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/migrations/op_set_unique.go
```diff
@@ -150,6 +150,9 @@ func (o *OpSetUnique) Validate(ctx context.Context, s *schema.Schema) error {
 	if o.Name == "" {
 		return FieldRequiredError{Name: "name"}
 	}
+	if err := validateName(o.Column); err != nil {
+		return err
+	}
```
andrew-farries, line 155: I don't think it makes sense to check the length of the column name on the operation here as it must already exist in the database. The check can be removed.

pkg/roll/execute.go
```diff
@@ -112,6 +112,9 @@ func (m *Roll) StartDDLOperations(ctx context.Context, migration *migrations.Mig
 func (m *Roll) ensureViews(ctx context.Context, schema *schema.Schema, version string) error {
 	// create schema for the new version
 	versionSchema := VersionedSchemaName(m.schema, version)
+	if err := migrations.ValidateName(versionSchema); err != nil {
+		return err
+	}
```
andrew-farries, line 117: This is too late to check the length of the new version schema - at this point a potentially lengthy migration start operation has already taken place so it's a poor user experience for the migration to fail with a validation check at this stage. I'd remove the check here and if we need it we can add a check earlier in the migration start process as a separate PR.

andrew-farries (changes requested):

andrew-farries:
Bump @0del 👋 

This PR is almost ready to go; it just needs some removals.

PR #280, opened 2024-02-07
Add SQL `onComplete` flag, allow `sql` migration to run with others

exekias:
This change updates the `sql` operation to:

* Allow for a new `onComplete` flag, that will make it run on the Complete phase, rather than doing it during Start (default behavior).
* Allows for `sql` operations next to others in the same migration. We added this limitation to ensure this operation doesn't affect others, especially around schema state management.

Having `sql` next to other operations has proven convenient sometimes, by adding `onComplete` flag, we can allow for these migrations to run and rely on views recreation based on the final state.


schema.json
```diff
@@ -312,6 +312,10 @@
         "up": {
           "description": "SQL expression for up migration",
           "type": "string"
+        },
+        "onComplete": {
+          "description": "SQL expression will run on complete step (rather than on start)",
+          "type": "boolean"
```
andrew-farries, line 318: This field can be set `"default": false`, then the generated field is just a `bool` rather than a `*bool` making it easier to work with.

pkg/migrations/op_raw_sql.go
```diff
@@ -36,11 +40,26 @@ func (o *OpRawSQL) Validate(ctx context.Context, s *schema.Schema) error {
...
-// this operation is isolated, cannot be executed with other operations
-func (o *OpRawSQL) IsIsolated() {}
+func (o *OpRawSQL) onComplete() bool {
+	if o.OnComplete != nil {
+		return *o.OnComplete
+	}
+	return false
+}
+
+// this operation is isolated when executed on start, cannot be executed with other operations
```
andrew-farries, line 57: I think the comment for this method can be moved to the interface.

pkg/roll/execute.go
```diff
@@ -179,7 +207,9 @@ func (m *Roll) createView(ctx context.Context, version, name string, table schem
 	}
 
 	_, err := m.pgConn.ExecContext(ctx,
-		fmt.Sprintf("CREATE OR REPLACE VIEW %s.%s %s AS SELECT %s FROM %s",
+		fmt.Sprintf("BEGIN; DROP VIEW IF EXISTS %s.%s; CREATE VIEW %s.%s %s AS SELECT %s FROM %s; COMMIT",
```
andrew-farries, line 210: why the change from `create or replace` to `drop and create`?

docs/README.md
```diff
@@ -1072,6 +1072,21 @@ Example **raw SQL** migrations:
...
+  "sql": {
+    "up": "SQL expression",
+    "onComplete": true
+  }
+}
+```
+
+The `onComplete` flag will make this operation run the `up` expression on the complete phase (instead of the default, which is to run it on the start phase).
+
+`onComplete` flag is incompatible with `down` expression, as `pgroll` does not support running rollback after complete was executed.
```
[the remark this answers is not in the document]
exekias: There could be a change to the schema in an `onComplete` migration, it's just we cannot undo it cause the migration is marked as done after complete. We only allow to rollback (and hence execute SQL down) migrations while they are active.

pkg/migrations/migrations.go
```diff
@@ -35,12 +35,12 @@ type Operation interface {
 // IsolatedOperation is an operation that cannot be executed with other operations
 // in the same migration
 type IsolatedOperation interface {
-	IsIsolated()
+	IsIsolated() bool
```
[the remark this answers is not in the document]
exekias: I kind of wanted to avoid having all operations implementing this, it increases the knowledge you need to write an op, while so far this is the only op needing it for real. That's why using duck typing here feels better to me, WDYT?

pkg/roll/execute.go
```diff
@@ -49,12 +49,15 @@ func (m *Roll) Start(ctx context.Context, migration *migrations.Migration, cbs .
...
-		if _, ok := op.(migrations.RequiresSchemaRefreshOperation); ok {
-			// refresh schema
-			newSchema, err = m.state.ReadSchema(ctx, m.schema)
-			if err != nil {
-				return fmt.Errorf("unable to refresh schema: %w", err)
+		// refresh schema when the op is isolated and requires a refresh (for example raw sql)
+		// we don't want to refresh the schema if the operation is not isolated as it would
+		// override changes made by other operations
+		if refreshOp, ok := op.(migrations.RequiresSchemaRefreshOperation); ok && refreshOp.RequiresSchemaRefresh() {
+			if isolatedOp, ok := op.(migrations.IsolatedOperation); ok && isolatedOp.IsIsolated() {
```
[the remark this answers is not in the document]
exekias: Main reason is: that it could be non-isolated & require schema refresh. That's the case for migrations with `onComplete`.

Migrations that are not isolated & require a schema refresh on `Start` phase would surely break others, as it would override the resulting schema from the start ops, by refreshing it afterwards. I tried to capture this in the comment

pkg/roll/execute.go
exekias: I found out that changes adding new columns to an existing view (because of a SQL op), could fail to `create or replace` the view, as it requires all columns to be in the same order. Since this is equivalent and doesn't fail I went for it

pkg/jsonschema/testdata/sql-4.txtar
```diff
@@ -0,0 +1,19 @@
+This is a valid 'sql' migration.
+It specifies `up`, `down` and `on_complete
```
andrew-farries, line 2:
```suggestion
It specifies `up`, `down` and `on_complete`
```

docs/README.md
```diff
@@ -1068,9 +1068,27 @@ A raw SQL operation runs arbitrary SQL against the database. This is intended as
...
+
+
+```json
+{
+  "sql": {
+    "up": "SQL expression",
+    "onComplete": true
+  }
+}
+```
```
andrew-farries, line 1090: This block should move upwards to before the examples too.

docs/README.md
```diff
@@ -1068,9 +1068,27 @@ A raw SQL operation runs arbitrary SQL against the database. This is intended as
 }
 ```
 
+By default, `sql` operation cannot run together with other operations in the same migration. This is to ensure pgroll can correctly track the state of the database. However, it is possible to run `sql` operation together with other operations by setting the `onComplete` flag to `true`.
```
andrew-farries, line 1071:
```suggestion
By default, a `sql` operation cannot run together with other operations in the same migration. This is to ensure pgroll can correctly track the state of the database. However, it is possible to run a `sql` operation together with other operations by setting the `onComplete` flag to `true`.
```

andrew-farries (approved):

Issue #283, opened 2024-02-21
Improve locking around pgroll initialization


andrew-farries:
Two processes calling `roll.Init` simultaneously have a chance of a deadlock:

https://github.com/xataio/pgroll/blob/main/pkg/roll/roll.go#L88-L90

Advisory locking could be a solution here.

PR #285, opened 2024-02-22
Make state initialization concurrency safe

andrew-farries:
Make `pgroll` state initialization concurrency safe by using Postgres advisory locking to ensure at most one connection can initialize at at time.

See docs on Postgres advisory locking:
* https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS
* https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS

Closes https://github.com/xataio/pgroll/issues/283

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -354,11 +354,26 @@ func New(ctx context.Context, pgURL, stateSchema string) (*State, error) {
 }
 
 func (s *State) Init(ctx context.Context) error {
-	// ensure pgroll internal tables exist
-	// TODO: eventually use migrations for this instead of hardcoding
-	_, err := s.pgConn.ExecContext(ctx, fmt.Sprintf(sqlInit, pq.QuoteIdentifier(s.schema)))
+	tx, err := s.pgConn.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
 
-	return err
+	// Try to obtain an advisory lock
+	const key int64 = 0x2c03057fb9525b
+	_, err = tx.ExecContext(ctx, "SELECT pg_advisory_xact_lock($1)", key)
+	if err != nil {
+		return err
+	}
+
+	// Perform pgroll state initialization
+	_, err = tx.ExecContext(ctx, fmt.Sprintf(sqlInit, pq.QuoteIdentifier(s.schema)))
+	if err != nil {
+		return err
+	}
+
+	return tx.Commit()
 }
 
 func (s *State) Close() error {
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -360,7 +360,9 @@ func (s *State) Init(ctx context.Context) error {
 	}
 	defer tx.Rollback()
 
-	// Try to obtain an advisory lock
+	// Try to obtain an advisory lock.
+	// The key is an arbitrary number, used to distinguish the lock from other locks.
+	// The lock is automatically released when the transaction is committed or rolled back.
 	const key int64 = 0x2c03057fb9525b
 	_, err = tx.ExecContext(ctx, "SELECT pg_advisory_xact_lock($1)", key)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/state/state.go
```diff
@@ -354,11 +354,26 @@ func New(ctx context.Context, pgURL, stateSchema string) (*State, error) {
...
-	_, err := s.pgConn.ExecContext(ctx, fmt.Sprintf(sqlInit, pq.QuoteIdentifier(s.schema)))
+	tx, err := s.pgConn.Begin()
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
 
-	return err
+	// Try to obtain an advisory lock
+	const key int64 = 0x2c03057fb9525b
```
eminano, line 364: Adding a comment here to explain this is a (randomly picked) unique int64 key for access to the pgroll initialisation resource that will release automatically at the end of the transaction would be good for readability/understanding when not familiar with advisory locks.

eminano (approved):

pkg/state/state.go
exekias: ++ also making it a constant would make sense

pkg/state/state.go
```diff
@@ -354,11 +354,26 @@ func New(ctx context.Context, pgURL, stateSchema string) (*State, error) {
 }
 
 func (s *State) Init(ctx context.Context) error {
-	// ensure pgroll internal tables exist
-	// TODO: eventually use migrations for this instead of hardcoding
-	_, err := s.pgConn.ExecContext(ctx, fmt.Sprintf(sqlInit, pq.QuoteIdentifier(s.schema)))
+	tx, err := s.pgConn.Begin()
```
exekias, line 357: do we really need a transaction? I'm wondering if this could cause locking for other table modifications, for instance if some triggers are using our functions

andrew-farries: I was unable to get the same liveness guarantees using a session level lock (`pg_advisory_lock`) as I was using the transaction level locks. Using the diff below and the test program from [this gist](https://gist.github.com/andrew-farries/883a9f314834067ca3301548844c7d1b) I was able to observe instances where all goroutines would become blocked or where Postgres would report deadlocks.

```diff
diff --git a/pkg/state/state.go b/pkg/state/state.go
index 79f3597..9eb1248 100644
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -354,28 +354,23 @@ func New(ctx context.Context, pgURL, stateSchema string) (*State, error) {
 }
 
 func (s *State) Init(ctx context.Context) error {
-	tx, err := s.pgConn.Begin()
-	if err != nil {
-		return err
-	}
-	defer tx.Rollback()
-
 	// Try to obtain an advisory lock.
 	// The key is an arbitrary number, used to distinguish the lock from other locks.
 	// The lock is automatically released when the transaction is committed or rolled back.
 	const key int64 = 0x2c03057fb9525b
-	_, err = tx.ExecContext(ctx, "SELECT pg_advisory_xact_lock($1)", key)
+	_, err := s.pgConn.ExecContext(ctx, "SELECT pg_advisory_lock($1)", key)
 	if err != nil {
 		return err
 	}
 
+	defer func() {
+		s.pgConn.ExecContext(ctx, "SELECT pg_advisory_unlock($1)", key)
+	}()
+
 	// Perform pgroll state initialization
-	_, err = tx.ExecContext(ctx, fmt.Sprintf(sqlInit, pq.QuoteIdentifier(s.schema)))
-	if err != nil {
-		return err
-	}
+	_, err = s.pgConn.ExecContext(ctx, fmt.Sprintf(sqlInit, pq.QuoteIdentifier(s.schema)))
 
-	return tx.Commit()
+	return err
 }
 
 func (s *State) Close() error {

```

exekias (approved):

PR #286, opened 2024-02-23
Relax add_column requirements on pk

exekias:
Primary keys are only required by pgroll when backfilling tables, relax checks in the `add_column` op so we only require a present PK if backfill will be needed (up function defined).


andrew-farries (commented):
We could add a test for this case to `TestAddColumnValidation`.

Update the existing `table must have a primary key on exactly one column` case to say it's only required if `up` is specified and add another case where the table can have > 1 column PK without `up` SQL.

andrew-farries (approved):

PR #287, opened 2024-02-26
Bump `golangci-lint` action to `v4`

andrew-farries:
Bump the `golangci-lint` action to `v4` and stop using the cache from `setup-go` as `golangci-lint` manages its own cache.

This removes all warnings from the summary page from workflow runs.


exekias (approved):
just a note, in the PR description you say V5 but the change is to V4

andrew-farries:
Thanks, I've updated the description.

PR #288, opened 2024-02-27
Relax backfill requirements so it works with unique columns

exekias:
UNIQUE NOT NULL columns should also work in order to perform backfills. This change relaxes the check on backfill requirements to use those columns if a primary key is not available.

Validation will still error out if no suitable column is found

Note: this also fixes the `unique` retrieval from schema, where we were failing to understand composed unique indices, resulting in columns flagged as unique where they weren't really unique.

### DIFF
--- a/pkg/migrations/backfill.go
+++ b/pkg/migrations/backfill.go
@@ -19,19 +19,18 @@ import (
 // 3. Update each row in the batch, setting the value of the primary key column to itself.
 // 4. Repeat steps 2 and 3 until no more rows are returned.
 func backfill(ctx context.Context, conn *sql.DB, table *schema.Table, cbs ...CallbackFn) error {
-	// Get the primary key column for the table
-	pks := table.GetPrimaryKey()
-	if len(pks) != 1 {
-		return errors.New("table must have a single primary key column")
+	// get the backfill column
+	pk := getBackfillColumn(table)
+	if pk == nil {
+		return BackfillNotPossible{Table: table.Name}
 	}
-	pk := pks[0]
 
 	// Create a batcher for the table.
 	b := batcher{
-		table:     table,
-		pkColumn:  pk,
-		lastPK:    nil,
-		batchSize: 1000,
+		table:          table,
+		backfillColumn: pk,
+		lastValue:      nil,
+		batchSize:      1000,
 	}
 
 	// Update each batch of rows, invoking callbacks for each one.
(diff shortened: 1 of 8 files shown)


pkg/migrations/backfill.go
```diff
@@ -19,19 +19,18 @@ import (
...
-	pk := pks[0]
 
 	// Create a batcher for the table.
 	b := batcher{
-		table:     table,
-		pkColumn:  pk,
-		lastPK:    nil,
-		batchSize: 1000,
+		table:          table,
+		backfillColumn: backFillColumn,
```
andrew-farries, line 31: Maybe `identityColumn` is a better name? `backfillColumn` makes me think it's the column that's going to be backfilled.

pkg/migrations/backfill.go
```diff
@@ -51,11 +50,38 @@ func backfill(ctx context.Context, conn *sql.DB, table *schema.Table, cbs ...Cal
...
+	// If there is no primary key, look for a unique not null column
+	for _, col := range table.Columns {
+		if col.Unique && !col.Nullable {
+			return &col
+		}
+	}
+
+	// no suitable column found
+	return nil
+}
```
andrew-farries, line 78: These two could be methods on `table` and moved alongside the other schema methods like `GetPrimaryKey`.

pkg/migrations/op_add_column_test.go
```diff
@@ -674,7 +766,7 @@ func TestAddColumnValidation(t *testing.T) {
 					},
 				},
 			},
-			wantStartErr: migrations.InvalidPrimaryKeyError{Table: "orders", Fields: 2},
+			wantStartErr: migrations.BackfillNotPossible{Table: "orders"},
```
andrew-farries, line 769: I think another test case is needed here to test that an operation that requires backfill is valid if it doesn't have a PK but does have a unique, not null column.

pkg/migrations/backfill.go
exekias: I thought about it, but the domain knowledge of backfill is in this package and not that one, so it felt a bit weird

pkg/migrations/op_add_column_test.go
exekias: I added `add column with up sql and no pk` for this

exekias: ah sorry, got it now, adding it 🙏

exekias: pushed [a28aec0](https://github.com/xataio/pgroll/pull/288/commits/a28aec0fc4e8a95981f710d2be00c50ad9aa27d2)

andrew-farries (approved):

PR #289, opened 2024-02-28
Separate DDL operations from DML (backfills) on migration start

andrew-farries:
Separate the DDL operations required during migration start from the DML operations (backfills). Complete all DDL steps before starting DML.

Each `Start` operation returns the name of the table that requires backfill, if any. After all operations have started, backfills are run on each table that requires one.

Separating DDL and DML during migration start like this will allow for setting Postgres options that should apply only for the DDL phase of migration start (as in #290).

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -17,7 +17,8 @@ type Operation interface {
 	// Start will apply the required changes to enable supporting the new schema
 	// version in the database (through a view)
 	// update the given views to expose the new schema version
-	Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) error
+	// Returns the table that requires backfilling, if any.
+	Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error)
 
 	// Complete will update the database schema to match the current version
 	// after calling Start.
--- a/pkg/migrations/op_add_column.go
+++ b/pkg/migrations/op_add_column.go
@@ -15,31 +15,32 @@ import (
 
 var _ Operation = (*OpAddColumn)(nil)
 
-func (o *OpAddColumn) Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) error {
+func (o *OpAddColumn) Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error) {
 	table := s.GetTable(o.Table)
 
 	if err := addColumn(ctx, conn, *o, table); err != nil {
-		return fmt.Errorf("failed to start add column operation: %w", err)
+		return nil, fmt.Errorf("failed to start add column operation: %w", err)
 	}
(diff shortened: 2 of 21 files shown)


pkg/migrations/op_add_column.go
```diff
@@ -52,18 +53,16 @@ func (o *OpAddColumn) Start(ctx context.Context, conn *sql.DB, stateSchema strin
...
-			return fmt.Errorf("failed to create trigger: %w", err)
-		}
-		if err := backfill(ctx, conn, table, cbs...); err != nil {
-			return fmt.Errorf("failed to backfill column: %w", err)
+			return nil, fmt.Errorf("failed to create trigger: %w", err)
 		}
+		tableToBackfill = table
 	}
 
 	table.AddColumn(o.Column.Name, schema.Column{
```
exekias, line 61: wondering if this could cause issues, we return a pointer to the table to backfill, then we add the new column to it. We are effectively adding a column here, and it could be a NOT NULL UNIQUE column, so backfill could potentially decide to use this one to backfill the very same column 😇 , which would create issues. Wondering how this can be solved, probably backfill can be smart enough to skip the backfilled column when choosing the identity

pkg/migrations/migrations.go
```diff
@@ -17,7 +17,8 @@ type Operation interface {
 	// Start will apply the required changes to enable supporting the new schema
 	// version in the database (through a view)
 	// update the given views to expose the new schema version
-	Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) error
+	// Returns the table that requires backfilling, if any.
+	Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error)
```
exekias, line 21: wondering about the API here, not a big deal but returning a table can be strange unless you read this comment. We could consider creating a new struct to return here, mostly for DX reasons, thoughts?

exekias (commented):
🚀 love to see this! it will actually speed up some migrations and provide better control & info over the migration progress! I left a few comments

pkg/migrations/op_add_column.go
andrew-farries: It's a good point about adding a `NOT NULL`, `UNIQUE` column here and then having `Backfill` decide to use it as the identity column for backfilling.

I don't think it's going to be a big issue in practice as `Backfill` will prefer to use a PK if one exists on the table.

I think it should be possible to have `Backfill` decline to use columns that have a `__pgroll_new` name as the identity. I can do that in a follow up PR if that works for you.

pkg/migrations/migrations.go
andrew-farries: Is there a better alternative than a single-field struct? I'm not sure that provides a cleaner API or better DX.

exekias: none than I can think of, I'm happy to iterate on this as we go

exekias (approved):

pkg/migrations/op_add_column.go
exekias: approved but please don't forget about this, I believe it's possible to get into an infinite loop with the current code

PR #293, opened 2024-03-01
Add `rename_constraint` operation

SferaDev:
In the frontend when renaming a column we were adding `OpRawSQL` operations that `ALTER TABLE "${tableName}" RENAME CONSTRAINT "${oldName}" TO "${newName}";` to keep the constraints up to date for the data api to like them. Proposing to expose a new operation for that use case.

Before merging, an example should be added for the new operation.

### DIFF
--- a/pkg/migrations/op_rename_constraint.go
+++ b/pkg/migrations/op_rename_constraint.go
@@ -14,8 +14,7 @@ import (
 var _ Operation = (*OpRenameConstraint)(nil)
 
 func (o *OpRenameConstraint) Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error) {
-	table := s.GetTable(o.Table)
-	table.RenameConstraint(o.From, o.To)
+	// no-op
 	return nil, nil
 }
 
--- a/pkg/migrations/op_rename_constraint.go
+++ b/pkg/migrations/op_rename_constraint.go
@@ -1,0 +1,48 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package migrations
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/schema"
+)
+
+var _ Operation = (*OpRenameConstraint)(nil)
+
+func (o *OpRenameConstraint) Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error) {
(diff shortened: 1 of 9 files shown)


pkg/schema/schema.go
```diff
@@ -201,6 +201,21 @@ func (t *Table) RenameColumn(from, to string) {
...
+		delete(t.CheckConstraints, from)
+	}
+	if _, ok := t.UniqueConstraints[from]; ok {
+		t.UniqueConstraints[to] = t.UniqueConstraints[from]
+		delete(t.UniqueConstraints, from)
+	}
+	if _, ok := t.ForeignKeys[from]; ok {
+		t.ForeignKeys[to] = t.ForeignKeys[from]
+		delete(t.ForeignKeys, from)
+	}
```
SferaDev, line 216: I think it's fine to rename the different constraint places we have in the schema, but this is the part I had more doubts. Please double check 😄

andrew-farries: This function isn't needed as there is no benefit to renaming the constraint in the virtual schema that gets built during migration start.

Its only purpose is to be able generate views with columns mapped to the right columns on the underlying table. Any constraints on the virtual schema aren't used for anything.

pkg/migrations/op_rename_constraint.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/schema"
+)
+
+var _ Operation = (*OpRenameConstraint)(nil)
+
+func (o *OpRenameConstraint) Start(ctx context.Context, conn *sql.DB, stateSchema string, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error) {
+	table := s.GetTable(o.Table)
+	table.RenameConstraint(o.From, o.To)
```
andrew-farries, line 18: As described in the other comment, this can be removed and the `Start` operation can be a no-op.

Drop index is another example of a similar kind of operation with a no-op `Start`.

andrew-farries (commented):
Looks good 👍 Thanks for doing it 🙏 

We also need:
* An example migration in `/examples` . Either add a migration to create a table with a constraint then another migration to rename it, or just add a migration to rename one of the constraints created in an existing example.
* Update the `docs/README.md` file operations reference section.

andrew-farries (commented):
I think the example can be:

```json
{
  "name": "33_rename_check_constraint",
  "operations": [
    {
      "rename_constraint": {
        "table": "people",
        "from": "name_length",
        "to": "name_length_check"
      }
    }
  ]
}
```

to rename one of the constraints created by an earlier migration.

andrew-farries (approved):

PR #294, opened 2024-03-01
Format `schema.json` on generate

SferaDev:
Add rule to format JSON Schema with prettier

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: generate
 generate:
 	# Format JSON schema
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,5 +1,9 @@
 .PHONY: generate
 generate:
+	# Format JSON schema
+	docker run --rm -v $$PWD/schema.json:/mnt/schema.json node npx prettier /mnt/schema.json --parser json --tab-width 2 --single-quote --trailing-comma all --no-semi --arrow-parens always --print-width 120 > schema.json.tmp
```
andrew-farries, line 4: The `node` image is over 1GB. There are other. smaller dockerized versions of `prettier` around like [this one](https://hub.docker.com/r/jauderho/prettier), but nothing official.

Using `node:alpine` here also works and the image is much smaller.

andrew-farries (approved):

PR #296, opened 2024-03-01
Add support for `onDelete` reference

### LINKED ISSUE #221: Support setting up ON DELETE on foreign keys
Currently there's not a way to set up ON DELETE options when setting up reference columns (foreign keys).

Specifically, ON DELETE CASCADE can be an extremely powerful tool for powering scenarios such as right to be forgotten requests. While setting this up can be accomplished through raw sql migrations, it would be ideal to have this as a setting on the reference itself.

SferaDev:
Fixes #221 

@andrew-farries I'm not sure if you finally implemented it, If you have something in the works please close this PR as yours will likely be better than mine.

The `enum` parsing could be way better, but `omissis/go-jsonschema` doesn't generate much without going all-in with their validator, and even with that it's behind `enumer` or `validator`. I decided to manually check for the values, but I'd be fine with not checking it at all or adding some better enum validator.

### DIFF
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -173,3 +173,12 @@ type InvalidReplicaIdentityError struct {
 func (e InvalidReplicaIdentityError) Error() string {
 	return fmt.Sprintf("replica identity on table %q must be one of 'NOTHING', 'DEFAULT', 'INDEX' or 'FULL', found %q", e.Table, e.Identity)
 }
+
+type InvalidEnumValueError struct {
+	Property string
+	Value    string
+}
+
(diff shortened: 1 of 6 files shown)


SferaDev:
Closing in favour of #297

andrew-farries:
Hi @SferaDev, thanks for doing this 🙏  . I was also working on this and have opened https://github.com/xataio/pgroll/pull/297 to do it.

It's basically the same as what we have here but also updates examples and the docs too. Mind if we close this?

andrew-farries:
haha, beat me to it 😆

PR #305, opened 2024-03-05
Fix duplicate inferred migrations when dropping columns outside of a migration

### LINKED ISSUE #304: Dropping columns results in duplicate `inferred` migrations
Dropping columns outside of a `pgroll` migration results in a duplicate `inferred` migration in the `pgroll.migrations` table.

**To reproduce**:

In a fresh Postgres instance:

* initialize `pgroll`:
```bash
pgroll init
```

* create a table
```sql
create table foo(a serial primary key, b text)
```

* drop a column:
```sql
alter table foo drop column b
```

* view the `pgroll.migrations` table:
```
select * from pgroll.migrations order by created_at desc
```

**Actual behaviour**:

The `pgroll.migrations` table contains two entries for the inferred column drop:

```
+--------+--------------------+-------------------------------------------------------------------------------------------------------------------+----------------------------+----------------------------+--------------------+------+-------------------------------------------------------------------------------------->
| schema | name               | migration                                                                                                         | created_at                 | updated_at                 | parent             | done | resulting_schema                                                                     >
|--------+--------------------+-------------------------------------------------------------------------------------------------------------------+----------------------------+----------------------------+--------------------+------+-------------------------------------------------------------------------------------->
| public | sql_0adcbadf878e88 | {"name": "sql_0adcbadf878e88", "operations": [{"sql": {"up": "alter table foo drop column b"}}]}                  | 2024-03-05 10:49:56.732626 | 2024-03-05 10:49:56.732626 | sql_a75d0c69f0cfe2 | True | {"name": "public", "tables": {"foo": {"oid": "16415", "name": "foo", "columns": {"a":>
| public | sql_ba9fefdb42a66c | {"name": "sql_ba9fefdb42a66c", "operations": [{"sql": {"up": "alter table foo drop column b"}}]}                  | 2024-03-05 10:49:56.732626 | 2024-03-05 10:49:56.732626 | sql_0adcbadf878e88 | True | {"name": "public", "tables": {"foo": {"oid": "16415", "name": "foo", "columns": {"a":>
| public | sql_a75d0c69f0cfe2 | {"name": "sql_a75d0c69f0cfe2", "operations": [{"sql": {"up": "create table foo(a serial primary key, b text)"}}]} | 2024-03-05 10:49:48.964148 | 2024-03-05 10:49:48.964148 | <null>             | True | {"name": "public", "tables": {"foo": {"oid": "16415", "name": "foo", "columns": {"a":>
+--------+--------------------+-------------------------------------------------------------------------------------------------------------------+
```

**Expected behaviour**:

There is exacly one `inferred` migration corresponding to the column drop.

andrew-farries:
Ensure that only one `inferred` migration is created in the `pgroll.migrations` table when a column is dropped outside of a migration.

From the Postgres [docs](https://www.postgresql.org/docs/current/event-trigger-definition.html):
> The sql_drop event occurs just before the ddl_command_end event trigger for any operation that drops database objects

This means that when the `raw_migration` function is run in response to `sql_drop` and `ddl_command_end`, duplicate entries will be created in `pgroll.migrations`; once as the function is run for `sql_drop` and again when it's run for `ddl_command_end`. 

Change the definition of the `pg_roll_handle_drop` event trigger to only run on those kinds of drops that won't result in duplicates when the `pg_roll_handle_ddl` trigger runs for the same change. `DROP TABLE` and `DROP VIEW` won't result in duplicate migrations because their schema can't be inferred by the `ddl_command_event` trigger because the object has already been dropped when the trigger runs.

Update the inferred migration tests with two new testcases covering dropping tables and columns.

Fixes #304

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -268,7 +268,7 @@ BEGIN
 		RETURN;
 	END IF;
 
-	IF tg_event = 'sql_drop' THEN
+	IF tg_event = 'sql_drop' and tg_tag != 'ALTER TABLE' THEN
 		-- Guess the schema from drop commands
 		SELECT schema_name INTO schemaname FROM pg_catalog.pg_event_trigger_dropped_objects() WHERE schema_name IS NOT NULL;
 
@@ -324,7 +324,6 @@ CREATE EVENT TRIGGER pg_roll_handle_ddl ON ddl_command_end
 DROP EVENT TRIGGER IF EXISTS pg_roll_handle_drop;
 CREATE EVENT TRIGGER pg_roll_handle_drop ON sql_drop
 	EXECUTE FUNCTION %[1]s.raw_migration();
-
 `
 
 type State struct {
--- a/pkg/state/state_test.go
+++ b/pkg/state/state_test.go
@@ -78,7 +78,7 @@ func TestInferredMigration(t *testing.T) {
 				},
 			},
 			{
-				name: "drop table",
+				name: "create/drop table",
 				sqlStmts: []string{
 					"CREATE TABLE table1 (id int)",
 					"DROP TABLE table1",
@@ -97,7 +97,7 @@ func TestInferredMigration(t *testing.T) {
 				},
 			},
 			{
-				name: "drop column",
+				name: "create/drop column",
 				sqlStmts: []string{
 					"CREATE TABLE table1 (id int, b text)",
 					"ALTER TABLE table1 DROP COLUMN b",
@@ -116,7 +116,7 @@ func TestInferredMigration(t *testing.T) {
 				},
 			},
 			{
-				name: "drop check constraint",
+				name: "create/drop check constraint",
 				sqlStmts: []string{
 					"CREATE TABLE table1 (id int, age integer, CONSTRAINT check_age CHECK (age > 0))",
 					"ALTER TABLE table1 DROP CONSTRAINT check_age",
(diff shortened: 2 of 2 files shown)


pkg/state/state.go
```diff
@@ -322,9 +322,8 @@ CREATE EVENT TRIGGER pg_roll_handle_ddl ON ddl_command_end
 	EXECUTE FUNCTION %[1]s.raw_migration();
 
 DROP EVENT TRIGGER IF EXISTS pg_roll_handle_drop;
-CREATE EVENT TRIGGER pg_roll_handle_drop ON sql_drop
+CREATE EVENT TRIGGER pg_roll_handle_drop ON sql_drop WHEN TAG IN ('DROP TABLE', 'DROP VIEW')
```
exekias, line 325: any chance we are missing other objects that could only be detected here? ie constraints, indices, triggers

exekias (commented):
change LGTM, but want to make sure we know we are covering all objects

pkg/state/state.go
andrew-farries: I've added several more test cases and changed how the trigger function definition so that `raw_migration` runs on all `sql_drop`s but doesn't get the schema for `ALTER TABLE DROP ...` as those will be handled when the function is called for `ddl_command_end`.

I think this way round reduces the chance of missing any operations.

exekias (approved):

PR #308, opened 2024-03-06
Respect foreign key `ON DELETE` in add column and create table operations

### LINKED ISSUE #306: Ensure foreign key `ON DELETE` option is respected for new columns
#297 added the ability to specify the `ON DELETE` behavior of a foreign key when altering an existing column.

The same setting needs to be respected for new columns (ie those created by add column and create table operations).

andrew-farries:
#297 added the ability to specify the `ON DELETE` behavior of a foreign key when altering an existing column.

Make it so that the `ON DELETE` property of the the FK is respected for FKs on new columns too (ie those created by add column and create table operations).

Fixes #306

### DIFF
--- a/pkg/migrations/errors.go
+++ b/pkg/migrations/errors.go
@@ -184,15 +184,13 @@ func (e InvalidReplicaIdentityError) Error() string {
 }
 
 type InvalidOnDeleteSettingError struct {
-	Table   string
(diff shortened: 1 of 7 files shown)


eminano (approved):

exekias (approved):
🚀

PR #315, opened 2024-03-11
Add `WithRawSQLURL` to configure an optional URl for raw sql operations

exekias:
This setting comes handy when you need to use a different connection string for raw SQL operations, as they may require some more security checks.

### DIFF
--- a/pkg/roll/execute.go
+++ b/pkg/roll/execute.go
@@ -4,6 +4,7 @@ package roll
 
 import (
 	"context"
+	"database/sql"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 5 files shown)


pkg/roll/execute.go
```diff
@@ -219,6 +220,18 @@ func (m *Roll) Rollback(ctx context.Context) error {
 	return nil
 }
 
+// connForOp returns the connection to use for the given operation
```
andrew-farries, line 223:
```suggestion
// connForOp returns the connection to use for the given operation.
```

pkg/roll/execute.go
```diff
@@ -219,6 +220,18 @@ func (m *Roll) Rollback(ctx context.Context) error {
...
+// if the operation is a raw SQL operation, it will use the rawSQLConn (if set)
```
andrew-farries, line 224:
```suggestion
// If the operation is a raw SQL operation, it will use the rawSQLConn (if set)
```

pkg/roll/execute_test.go
```diff
@@ -507,6 +508,78 @@ func TestMigrationHooksAreInvoked(t *testing.T) {
...
+		t.Cleanup(func() {
+			if err := st.Close(); err != nil {
+				t.Fatal(err)
+			}
+		})
+
+		db, err := sql.Open("postgres", connStr)
+		if err != nil {
+			t.Fatal(err)
+		}
```
andrew-farries, line 535: I think this test could be made shorter by moving more of this setup into `WithConnectionString`. `WithConnectionString` can create the state. The signature of the callback could be: `fn func(*state.State, *sql.DB, connStr string)`

andrew-farries (approved):

PR #318, opened 2024-03-12
Move example schema validation to its own job

andrew-farries:
Move the validation of the examples' JSON schema from a step in the `example` job into its own job.

In its previous location the examples were having their schema validated in each job in the matrix (pg version x public/non_public schema), making each job in the matrix take ~15s longer.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -172,7 +172,7 @@ jobs:
 
   release:
     runs-on: ubuntu-latest
-    needs: [test, lint, examples-schema-validation ,examples, license-check, type-generation, dead-code-check]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -166,7 +172,7 @@ jobs:
 
   release:
     runs-on: ubuntu-latest
-    needs: [test, lint, examples, license-check, type-generation, dead-code-check]
+    needs: [test, lint, examples-schema-validation ,examples, license-check, type-generation, dead-code-check]
```
exekias, line 175:
```suggestion
    needs: [test, lint, examples-schema-validation, examples, license-check, type-generation, dead-code-check]
```

PR #320, opened 2024-03-15
Use the connection defined by `WithRawSQLURL` for backfill operations

andrew-farries:
https://github.com/xataio/pgroll/pull/315 added the option to use a separate connection for raw SQL operations.

The same arguments about using a separate, more restrictive connection to run raw SQL operations also apply to running `up` and `down` SQL during backfill operations.

This PR changes the backfill operation to use the alternative connection if defined.

### DIFF
--- a/pkg/roll/execute.go
+++ b/pkg/roll/execute.go
@@ -232,6 +232,16 @@ func (m *Roll) connForOp(op migrations.Operation) *sql.DB {
 	return m.pgConn
 }
 
+// connForBackfills returns the connection to use for backfills. If a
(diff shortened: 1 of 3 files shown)


exekias (approved):

andrew-farries:
@exekias I don't think this PR makes sense, as the backfill will ultimately be performed by the code in the trigger.

Altering the connection used to call the `backfill` method won't affect what the trigger is able to run even if there is a proxy on the connection that starts the backfill process; the only code run by that connection directly is the `SELECT....` statements executed by the batcher.

I think the PR should be closed.

exekias:
SGTM!

PR #326, opened 2024-03-22
Add `Upper` and `Downer` interfaces to identify operations that create `up` and `down`  triggers

andrew-farries:
Add two new interfaces `Upper` and `Downer`, implemented by operations that create up and down triggers.

Consumers of `pgroll` may wish to identify operations that run arbitrary SQL via `up` and `down` triggers and pre-process them in such a way to ensure that the SQL is safe, or otherwise restricted.

With the new interfaces clients can do something like:

```go
  if up, ok := op.(migrations.Upper); ok {
    processedSQL := processSQL(up.UpSQL())
    up.SetUpSQL(processedSQL)
  }
  if down, ok := op.(migrations.Downer); ok {
    processedSQL := processSQL(down.DownSQL())
    down.SetDownSQL(processedSQL)
  }
}
```

to iterate over all operations in a migration and process the `up` or `down` SQL of each operation.

Update the `createTrigger` function to check that the operation creating the trigger implements either `Upper` or `Downer` according to the direction of the trigger. This guards against future operations creating triggers without implementing `Upper` or `Downer`.

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -46,6 +46,22 @@ type RequiresSchemaRefreshOperation interface {
 	RequiresSchemaRefresh()
 }
 
+// Upper is implemented by operations that define `up` SQL
+// Implementations should return the user-defined `up` SQL for the operation,
+// and not any default `up` SQL that may apply to the operation.
+type Upper interface {
+	UpSQL() string
+	SetUpSQL(string)
+}
+
+// Downer is implemented by operations that define `down` SQL
+// Implementations should return the user-defined `down` SQL for the operation,
+// and not any default `down` SQL that may apply to the operation.
+type Downer interface {
+	DownSQL() string
+	SetDownSQL(string)
+}
+
 type (
 	Operations []Operation
 	Migration  struct {
--- a/pkg/migrations/op_add_column.go
+++ b/pkg/migrations/op_add_column.go
@@ -182,6 +182,9 @@ func (o *OpAddColumn) Validate(ctx context.Context, s *schema.Schema) error {
 	return nil
 }
 
+func (o *OpAddColumn) UpSQL() string      { return o.Up }
+func (o *OpAddColumn) SetUpSQL(up string) { o.Up = up }
+
 func addColumn(ctx context.Context, conn *sql.DB, o OpAddColumn, t *schema.Table) error {
 	// don't add non-nullable columns with no default directly
 	// they are handled by:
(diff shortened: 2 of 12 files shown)


pkg/migrations/migrations.go
```diff
@@ -46,6 +46,22 @@ type RequiresSchemaRefreshOperation interface {
...
+	SetUpSQL(string)
+}
+
+// Downer is implemented by operations that define `down` SQL
+// Implementations should return the user-defined `down` SQL for the operation,
+// and not any default `down` SQL that may apply to the operation.
+type Downer interface {
+	DownSQL() string
+	SetDownSQL(string)
+}
```
exekias, line 63: with this approach we are kind of forcing any op to follow this pattern, so every single op is responsible of correctly implementing this. I'm wondering if we could be better by injecting a SQL transformer as a parameter to `roll.New`, that gets called by `trigger.go` only, WDYT? I believe it would be similar to the current hooks

andrew-farries: It depends how we think SQL transformation will be used. If we go with this approach of defining interfaces that make it easier to work with operations that define `up` or `down` SQL, it can be done by clients as a separate step, before a migration is run, eg an RPC like `SanitizeMigration` that returns a migration with all `up`/`down` SQL transformed or an error if restricted SQL is present.

Trggers are created part way through the DDL phase for an operation. Transforming SQL is a validation step that can fail and IMO we ought to do it before running any DDL. If we make it part of validation, then all operations will be responsible for implementing that correctly, just as they are responsible for implementing `Upper` and `Downer` with this approach. With this current approach and op that doesn't implement the interfaces at all will be blocked from creating a trigger.

I'd like to remove the repetition where each op with `up` or `down` has to implement these methods by embedding a struct that implements the interfaces into each such operation. However, our operation types are generated from the JSON schema which restricts how we are able to define them; I don't think it's possible to get the [go-jsonschema](https://github.com/omissis/go-jsonschema) package to emit embedded structs.

exekias: gotcha, having access to the up/down functions at validation time makes a lot of sense. Perhaps we could modify the interface only to expose these triggers through a known method? this way we can check the trigger at validation time, and use that same function from the start phase to setup the triggers (and modify the functions if needed)

andrew-farries: So yout think we should move SQL transformation into the `Validate` step for each operation? This would mean that validation could potentially alter the migration to be run, by rewriting the `up`/`down` SQL.

If we stick with the approach in this PR any SQL transformation can be done as a separate step before starting a migration. I think this has some benefits in terms of visibility for users in terms of what SQL will be run.

andrew-farries:
Closing this after some discussion in favour of an option-based approach.

PR #329, opened 2024-03-27
Add a `WithSQLTransformer` option to rewrite user-defined SQL in `up` and `down` triggers

andrew-farries:
Add a new `WithSQLTransformer` option to rewrite the user-defined SQL used to define `up` and `down` triggers. 

The intention is that the transformer be used to sanitize user-input SQL.

Transformers implement the following interface:

```go
type SQLTransformer interface {
	Transform(sql string) (string, error)
}
```

and are used by the `createTrigger` function to rewrite the `up` or `down` SQL before using it in the trigger function definition.

Later PRs will use the same transformer to rewrite the `up` and `down` values used in raw SQL migrations and column `DEFAULT` expressions.

### DIFF
--- a/pkg/migrations/migrations.go
+++ b/pkg/migrations/migrations.go
@@ -46,6 +46,16 @@ type RequiresSchemaRefreshOperation interface {
 	RequiresSchemaRefresh()
 }
 
+type SQLTransformer interface {
+	TransformSQL(sql string) (string, error)
+}
+
+type SQLTransformerFunc func(string) (string, error)
+
+func (fn SQLTransformerFunc) TransformSQL(sql string) (string, error) {
+	return fn(sql)
+}
+
 type (
 	Operations []Operation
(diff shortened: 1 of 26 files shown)


pkg/migrations/migrations.go
```diff
@@ -46,6 +46,16 @@ type RequiresSchemaRefreshOperation interface {
 	RequiresSchemaRefresh()
 }
 
+type SQLTransformer interface {
+	Transform(sql string) (string, error)
+}
+
+type SQLTransformerFunc func(string) (string, error)
+
+func (fn SQLTransformerFunc) Transform(sql string) (string, error) {
```
exekias, line 55: nit (feel free to ignore): `TransformSQL` may be a better name to avoid confusion

exekias (approved):
nice! 🚀

PR #331, opened 2024-03-28
Strengthen SQL transformation tests

andrew-farries:
Improve the existing tests around SQL transformation by adding a `MockSQLTransformer` type and using it in the existing SQL transformation tests. 

Also add several new testcases for transformation of raw SQL operations to ensure that when transformation fails the operation fails.

### DIFF
--- a/pkg/testutils/sql_transformer.go
+++ b/pkg/testutils/sql_transformer.go
@@ -25,8 +25,8 @@ func NewMockSQLTransformer(ts map[string]string) *MockSQLTransformer {
 // provided to the MockSQLTransformer. If the input SQL is not in the transformations
 // map, the input SQL is returned unchanged.
 func (s *MockSQLTransformer) TransformSQL(sql string) (string, error) {
-	out, ok := s.transformations[sql]
-	if !ok {
+	out, found := s.transformations[sql]
+	if !found {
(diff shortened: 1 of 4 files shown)


pkg/testutils/sql_transformer.go
```diff
@@ -0,0 +1,38 @@
...
+		transformations: ts,
+	}
+}
+
+// TransformSQL transforms the given SQL string according to the transformations
+// provided to the MockSQLTransformer. If the input SQL is not in the transformations
+// map, the input SQL is returned unchanged.
+func (s *MockSQLTransformer) TransformSQL(sql string) (string, error) {
+	out, ok := s.transformations[sql]
+	if !ok {
```
SferaDev, line 29: Not sure if naming convention to use `ok` for dictionary access, but felt a bit strange that if not ok, return input when reading the code.

```suggestion
	out, found := s.transformations[sql]
	if !found {
```

SferaDev (approved):

pkg/testutils/sql_transformer.go
andrew-farries: `ok` is a Go convention for the second return value of a map access, but here it's probably better to use `found` as you suggest.

PR #332, opened 2024-03-28
Rewrite column `DEFAULT`s using the SQL transformer

andrew-farries:
Use the SQL transformer added in #329 to rewrite or reject column `DEFAULT` values.

Column `DEFAULT` values are user-supplied SQL expressions that may need to be restricted or rewritten in some environments.

### DIFF
--- a/pkg/migrations/op_create_table.go
+++ b/pkg/migrations/op_create_table.go
@@ -15,10 +15,17 @@ import (
 var _ Operation = (*OpCreateTable)(nil)
 
 func (o *OpCreateTable) Start(ctx context.Context, conn *sql.DB, stateSchema string, tr SQLTransformer, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error) {
+	// Generate SQL for the columns in the table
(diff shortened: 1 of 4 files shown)


pkg/migrations/op_create_table.go
```diff
@@ -15,10 +15,17 @@ import (
 var _ Operation = (*OpCreateTable)(nil)
 
 func (o *OpCreateTable) Start(ctx context.Context, conn *sql.DB, stateSchema string, tr SQLTransformer, s *schema.Schema, cbs ...CallbackFn) (*schema.Table, error) {
+	// Generate SQL for the columns in the table
+	columnsSQL, err := columnsToSQL(o.Columns, tr)
```
SferaDev, line 19: [nit] I believe everywhere else we use `colSQL` naming and here we are using `columnsSQL`

SferaDev (approved):

pkg/migrations/op_create_table.go
andrew-farries: I think it's fine because those occurences are for a column (singular) in a helper function; here it's a plural in a 'top level' `Start` function.

PR #341, opened 2024-04-12
Allow rename-only 'alter-column' operations on unbackfillable columns

### LINKED ISSUE #340: Rename-only 'alter column' operations shouldn't be restricted to tables with PKs
Most 'alter column' sub-operations require that the table has a single column primary key or a unique, non-nullable column. This is required for backfilling.

'alter column' operations that have no sub-operations other than a rename should work on tables that don't meet this restriction because they don't require backfills.

This series of migrations should work:

```json
{
  "name": "01_create_table",
  "operations": [
    {
      "create_table": {
        "name": "products",
        "columns": [
          {
            "name": "name",
            "type": "varchar(255)"
          }
        ]
      }
    }
  ]
}
```

```json
{
  "name": "02_rename_column",
  "operations": [
    {
      "alter_column": {
        "table": "products",
        "column": "name",
        "name": "bananas"
      }
    }
  ]
}
```

but the second migration currently fails with:

```
Failed to start migration: migration is invalid: a backfill is required but table "products" doesn't have a single column primary key or a UNIQUE, NOT NULL column
```

andrew-farries:
Allow rename-only 'alter column' operations on unbackfillable columns. Rename operations don't require backfills so there is no reason to impose such a restriction.

Fixes #340

### DIFF
--- a/pkg/migrations/op_alter_column.go
+++ b/pkg/migrations/op_alter_column.go
@@ -189,10 +189,12 @@ func (o *OpAlterColumn) Validate(ctx context.Context, s *schema.Schema) error {
 		return ColumnDoesNotExistError{Table: o.Table, Name: o.Column}
 	}
 
-	// Ensure that the column has a primary key defined on exactly one column.
-	err := checkBackfill(table)
-	if err != nil {
-		return err
+	// If the operation requires backfills (ie it isn't a rename-only operation),
+	// ensure that the column meets the requirements for backfilling.
+	if !o.isRenameOnly() {
+		if err := checkBackfill(table); err != nil {
+			return err
+		}
 	}
 
 	// If the column is being renamed, ensure that the target column name does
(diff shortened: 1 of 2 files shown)


pkg/migrations/op_alter_column.go
```diff
@@ -189,9 +189,9 @@ func (o *OpAlterColumn) Validate(ctx context.Context, s *schema.Schema) error {
 		return ColumnDoesNotExistError{Table: o.Table, Name: o.Column}
 	}
 
-	// Ensure that the column has a primary key defined on exactly one column.
-	err := checkBackfill(table)
-	if err != nil {
+	// If the operation requires backfills (ie it isn't a rename-only operation),
+	// ensure that the column meets the requirements for backfilling.
+	if err := checkBackfill(table); !o.isRenameOnly() && err != nil {
```
exekias, line 194: should we avoid calling checkBackfill if we don't plan to backfill?

andrew-farries: Changed this so that we avoid the check entirely for rename only column alters.

exekias (approved):

PR #345, opened 2024-04-23
Support setting table and column comments to `NULL`

andrew-farries:
Build on #344 to allow removing column comments by setting them to `null`.

Make use of https://github.com/omissis/go-jsonschema/pull/220 and use the [nullable](https://github.com/oapi-codegen/nullable) package so that it's possible to distingush between a missing `comment` field and one that is explicitly set to `null`.

With https://github.com/omissis/go-jsonschema/pull/220 not being part of a release yet, use a custom build of `go-jsonschema`. It should be possible to switch back to the official release images once https://github.com/omissis/go-jsonschema/pull/220 is part of a release.

Without this change it becomes impossible to remove a comment from a column using the 'set comment' 'alter column' sub-operation (https://github.com/xataio/pgroll/pull/344).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,7 @@ jobs:
 
       - name: Validate example migrations against JSON schema
         run: |
-          npx -y ajv-cli --spec=draft2020 validate -s schema.json -d "./examples/*.json"
+          npx -y ajv-cli --spec=draft2020 validate --strict-schema=false -s schema.json -d "./examples/*.json"
 
   examples:
     name: 'examples (pg: ${{ matrix.pgVersion }}, schema: ${{ matrix.testSchema }})'


exekias:
uhm, I think the need for quotes is quite unexpected as a user. Wouldn't be enough to pass a `"comment": null` JSON value?

andrew-farries:
>  Wouldn't be enough to pass a "comment": null JSON value?

Passing `"comment": null` would mean that we can't easily distinguish between a `comment` field that is unset vs one that is explicitly set to `NULL`.  In an 'alter column' operation therefore, we wouldn't know whether to set the comment to `NULL` or leave it unchanged. This is a common problem with Go JSON unnmarshalling (see eg [here](https://www.jvt.me/posts/2024/01/09/go-json-nullable/)).

`pgroll` already requires quoted literal strings when setting column `DEFAULT` values; here the user may set the column `DEFAULT` to some function call (eg `now()` for timestamp columns)  so `pgroll` can't assume that all `DEFAULT`s should be quoted so it is left to the user to do so.

In the case of comments, the only two values allowed are literal strings or `NULL`. So arguably it's more surprising that non-`NULL` comments have to be single quoted strings than it is when setting a `DEFAULT` where arbitrary expressions are also allowed.

We have some options:
* Accept the changes in this PR and require `comment` fields to be single-quoted strings or a bare `NULL`.
* Close this PR and accept that comments can't be removed by alter column statements, only set to the empty string (removal of a comment could still be done with a raw SQL migration)
* Invest more time in coming up with a solution to allow distinguishing explicitly set to `NULL` comments vs missing fields (eg using the [nulllable package](https://pkg.go.dev/github.com/oapi-codegen/nullable#example-NewNullNullable)). This might be complicated by the fact that our operation types are generated from the JSON schema so doesn't allow us much control over the type of the `comment` field in an `OpAlterColumn`.

andrew-farries:
After some more experimentation, allowing `"comment": null` to remove a comment field will work as long as we can specify the type of the `OpAlterColumn.Comment` field as `Nullable[string]` (using the [nullable package](https://github.com/oapi-codegen/nullable)).

It requires an upstream change to [go-jsonschema](https://github.com/omissis/go-jsonschema) in order to emit the `nullable.Nullable` type correctly from the `schema.json` file.

This would mean users could specify either:
```
comment: null
```
or
```
comment: "some string"
```

without having to enclose `"some string"` in single quotes. This makes it inconsistent with how string literals are specified with `DEFAULT` values, but perhaps less surprising.

exekias:
I like this one! Wondering if we could do the same for others. roping @eemmiillyy @SferaDev for thoughts on the json schema

schema.json
```diff
@@ -159,7 +159,12 @@
         },
         "comment": {
           "description": "New comment on the column",
-          "type": "string"
+          "type": "string",
+          "goJSONSchema": {
+            "imports": ["github.com/oapi-codegen/nullable"],
+            "nillable": true,
+            "type": "nullable.Nullable[string]"
+          }
```
SferaDev, line 167: Ideally it would be something like this:

```suggestion
          "type": ["string", "null"],
```

andrew-farries: The problem with using `type: ["string", "null"]` is that it generates the Go type:

```go
Comment *string `json:"name"
```

which doesn't allow us to distinguish between the case where the  'alter column' `comment` field is specified and set to `null`, vs left unspecified.

See https://github.com/xataio/pgroll/pull/345#issuecomment-2071930676

Using `nullable.Nullable` allows us to distinguish the 'set tno null' and the 'unspecified' cases.

SferaDev: Fair point, but from the user land perspective (instead of go code) I would like to see that the field can be set to null. So I think both things should be set, right?

schema.json
```diff
@@ -159,7 +159,12 @@
         },
         "comment": {
           "description": "New comment on the column",
-          "type": "string"
+          "type": "string",
```
SferaDev, line 162:
```suggestion
          "type": ["string", "null"],
```

andrew-farries:
I think this is ready now. The upstream change to `go-jsonschema` has landed https://github.com/omissis/go-jsonschema/pull/220.

I've updated the PR description:

---
Build on #344 to allow removing column comments by setting them to `null`.

Make use of https://github.com/omissis/go-jsonschema/pull/220 and use the [nullable](https://github.com/oapi-codegen/nullable) package so that it's possible to distingush between a missing `comment` field and one that is explicitly set to `null`.

With https://github.com/omissis/go-jsonschema/pull/220 not being part of a release yet, use a custom build of `go-jsonschema`. It should be possible to switch back to the official release images once that PR is part of a release.

Without this change it becomes impossible to remove a comment from a column using the 'set comment' 'alter column' sub-operation (https://github.com/xataio/pgroll/pull/344).

andrew-farries:
> Wondering if we could do the same for others

What do you mean by this @exekias? I don't think the same approach makes sense for setting `default` values for the reasons described in https://github.com/xataio/pgroll/pull/345#issuecomment-2071930676.

It makes sense for `comment` because there are only two allowed values: string literals or `null`.

exekias (approved):

andrew-farries:
We also need to approve the PR on which this one is based: https://github.com/xataio/pgroll/pull/344

Issue #348, opened 2024-05-02
Altering a column type attempts to duplicate incompatible defaults and constraints


andrew-farries:
Columns that are duplicated  as part of an 'alter column' migration that changes the column type should not attempt to duplicate any `DEFAULT`s or `CHECK` constraints that are not compatible with the new column type.

Given this pair of migrations:

```json
{
  "name": "01_create_table",
  "operations": [
    {
      "create_table": {
        "name": "products",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "age",
            "type": "text",
            "check": {
              "name": "length_check",
              "constraint": "LENGTH(\"age\") <= 3"
            }
          }
        ]
      }
    }
  ]
}
```

```json
{
  "name": "02_alter_column",
  "operations": [
    {
      "alter_column": {
        "table": "products",
        "column": "age",
        "type": "bigint",
        "default": "0",
        "up": "CAST(age AS integer)",
        "down": "CAST(age AS text)"
      }
    }
  ]
}
```

will fail because the `length_check` constraint can't be applied to the new column type. A similar problem occurs if the column has a string default and the column type is changed to an integer.

As a first pass, it could be acceptable to simply drop defaults and constraints from columns that have their type changed; users can re-set them in the same 'alter column'  operation that changes the type if they are intended to be preserved.

Ideally the default/constraints should be dropped if they are not compatible with the new column type and preserved otherwise.

PR #349, opened 2024-05-03
Dont duplicate `CHECK` constraints and `DEFAULT`s when altering column type

andrew-farries:
When a column is duplicated with a new type, check constraints and defaults defined on the column *may* be incompatible with the new type.

In this case column duplication should not fail, but rather the incompatible constraints and defaults should be ignored and not be duplicated to the new column.

This PR changes column duplication to ignore errors on `DEFAULT` and `CHECK` constraint duplication that look as though they are caused by a change of column type.

Fixes https://github.com/xataio/pgroll/issues/348

### DIFF
--- a/pkg/migrations/duplicate.go
+++ b/pkg/migrations/duplicate.go
@@ -4,6 +4,7 @@ package migrations
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"slices"
 	"strings"
@@ -22,6 +23,11 @@ type Duplicator struct {
 	withType          string
 	withoutConstraint string
 }
+
+const (
+	dataTypeMismatchErrorCode  pq.ErrorCode = "42804"
+	undefinedFunctionErrorCode pq.ErrorCode = "42883"
+)
 
 // NewColumnDuplicator creates a new Duplicator for a column.
 func NewColumnDuplicator(conn db.DB, table *schema.Table, column *schema.Column) *Duplicator {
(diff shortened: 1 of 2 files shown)


exekias (approved):
I understand this doesn't affect other constraints, like UNIQUE or NOT NULL right?

andrew-farries:
>  I'm wondering if this will be unexpected for compatible types, ie when I cast from integer to int8 or float 🤔

That was my concern too. 

If we had done this before supporting multiple sub-operations in one alter column statement it would probably not have been the right decision.

Now however if you change type to a compatible type you can recreate the `DEFAULT` and `CHECK` in the same alter as the type change:

```json
{
  "name": "35_alter_column_multiple",
  "operations": [
    {
      "alter_column": {
        "table": "users",
        "column": "age",
        "type": "bigint",
        "default": "0",
        "check": {
          "name": "age_check",
          "constraint": "age > 21"
        },
        "up": "...",
        "down": "..."
      }
    }
  ]
}
```

which mitigates the problem, although it might still be unexpected that you have to do this.

Ideally we would be to try to preserve checks and defaults iff they are compatible with the new type.

andrew-farries:
We should preserve defaults and check constraints when they are compatible.

Putting this back into draft in order to do that.

andrew-farries:
Updated this to preserve `DEFAULT` and `CHECK` constraints when they are compatible, and ignore errors on duplication of `DEFAULT` and `CHECK` constraints when the error code suggests the failure is a result of column type change.

exekias (approved):

PR #365, opened 2024-07-01
Add `WithNoVersionSchemaForRawSQL` option

andrew-farries:
Add a new `WithNoVersionSchemaForRawSQL` option to control whether or not version schema should be created for raw SQL migrations.

With the option set, a raw SQL migration:
* Has no version schema created on migration start.
* Leaves the previous version schema in place on migration completion.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -70,22 +70,27 @@ LANGUAGE SQL
 STABLE;
 
 -- Get the name of the previous version of the schema, or NULL if there is none.
+-- This ignores previous versions for which no version schema exists, such as
+-- versions corresponding to inferred migrations.
 CREATE OR REPLACE FUNCTION %[1]s.previous_version(schemaname NAME) RETURNS text
 AS $$
-  WITH RECURSIVE find_ancestor AS (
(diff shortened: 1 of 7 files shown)


pkg/state/state.go
```diff
@@ -70,22 +70,27 @@ LANGUAGE SQL
 STABLE;
 
 -- Get the name of the previous version of the schema, or NULL if there is none.
+-- This ignores previous versions for which no version schema exists, such as
+-- versions corresponding to inferred migrations.
```
exekias, line 74: is this always the behavior? I wonder if this should be based on the setting

exekias (approved):

pkg/state/state.go
andrew-farries: It's equivalent to the old behaviour when the new setting is off. The only migrations that didn't create version schema were `inferred` migrations which were already filtered out by this function.

andrew-farries: The only use of this function is for cleaning up old version schema, for which this version of the function is fine. If we have a need to find the previous version, even if it doesn't have a version schema we can handle that then.

PR #369, opened 2024-07-09
Ignore duplicate inferred migrations having the same timestamp

andrew-farries:
Remove all but the last inferred migration in a group of inferred migrations that occur at the same timestamp with the same migration SQL.

Such inferred migrations are typically part of a batched transaction for which we want to ignore all but the effect of the last statement in the batch.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -311,6 +311,15 @@ BEGIN
 		RETURN;
 	END IF;
 
+	-- Remove any duplicate inferred migrations with the same timestamp for this
+	-- schema. We assume such migrations are multi-statement batched migrations
+	-- and we are only interested in the last one in the batch.
+	DELETE FROM %[1]s.migrations
+	WHERE schema = schemaname
+	AND created_at = current_timestamp
+	AND migration_type = 'inferred'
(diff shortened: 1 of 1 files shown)


pkg/state/state.go
```diff
@@ -311,6 +316,15 @@ BEGIN
...
 	END IF;
 
+	-- Remove any duplicate inferred migrations with the same timestamp for this
+	-- schema. We assume such migrations are multi-statement batched migrations
+	-- and we are only interested in the last one in the batch.
+	DELETE FROM %[1]s.migrations
+	WHERE schema = schemaname
+	AND created_at = current_timestamp
+	AND migration_type = 'inferred'
+	AND migration->'operations'->0->'sql'->>'up' = current_query();
```
exekias, line 326: Would it be safer to avoid inserting it at all?

exekias (approved):

pkg/state/state.go
andrew-farries: I think we need to insert so that we ensure we insert the final statement in the batch. Otherwise how would we know which was the last one?

PR #372, opened 2024-07-17
Minor attempt to clear up the behaviour in the documentation

dolezvo1:
I'm not sure what is the proper terminology for `public` vs `public_03_add_column` schemas. Major/main vs minor/specific? I think there definitely should be a consistent terminology, though.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -600,7 +600,7 @@ This completes the most recently started migration.
 
 Running `pgroll complete` when there is no migration in progress is a no-op.
 
-Completing a `pgroll` migration removes the previous schema version from the database (e.g. `public_02_create_table`), leaving only the latest version of the schema (e.g. `public_03_add_column`), available under both specific and main schema name (i.e. both `public_03_add_column` and `public`).
+Completing a `pgroll` migration removes the previous schema version from the database (e.g. `public_02_create_table`), leaving only the latest version of the schema (e.g. `public_03_add_column`). At this point, any temporary columns and triggers created on the affected tables in the `public` schema will also be cleaned up, leaving the table schema in its final state.
 
 :warning: Before running `pgroll complete` ensure that all applications that depend on the old version of the database schema are no longer live. Prematurely running `pgroll complete` can cause downtime of old application instances that depend on the old schema.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -574,7 +574,7 @@ $ pgroll start sql/03_add_column.json
 
 This starts the migration defined in the `sql/03_add_column.json` file.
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -572,9 +572,11 @@ The tables and functions in this schema store `pgroll`'s internal state and are
 $ pgroll start sql/03_add_column.json
 ```
 
-This starts the migration defined in the `sql/01_create_table.json` file.
+This starts the migration defined in the `sql/03_add_column.json` file.
 
-After starting a migration there will be two schema versions in the database; one for the old schema before the migration and one for the new version with the schema changes.
+After starting a migration there will be two schema versions in the database; one for the old schema before the migration (e.g. `public_02_create_table`) and one for the new version with the schema changes (e.g. `public_03_add_column`). At this point, the earlier version will be still available under the main schema name (e.g. `public`).
```
andrew-farries, line 577: I think it's important to make it clear that the versioned schema (like `public_02_create_table`) contain views on the tables in the `public` schema. Applications using schema managed by`pgroll` shouldn't connect directly to `public` but rather use the versioned views.

```suggestion
After starting a migration there will be two schema versions in the database; one for the old schema before the migration (e.g. `public_02_create_table`) and one for the new version with the schema changes (e.g. `public_03_add_column`).  Each schema contains views on the tables in the `public` schema.
```

docs/README.md
```diff
@@ -598,7 +600,7 @@ This completes the most recently started migration.
 
 Running `pgroll complete` when there is no migration in progress is a no-op.
 
-Completing a `pgroll` migration removes the previous schema version from the database, leaving only the latest version of the schema.
+Completing a `pgroll` migration removes the previous schema version from the database (e.g. `public_02_create_table`), leaving only the latest version of the schema (e.g. `public_03_add_column`), available under both specific and main schema name (i.e. both `public_03_add_column` and `public`).
```
andrew-farries, line 603: Another attempt to make it clear that the version schema contain views and the base schema (default `public`) contains the tables.

```suggestion
Completing a `pgroll` migration removes the previous schema version from the database (e.g. `public_02_create_table`), leaving only the latest version of the schema (e.g. `public_03_add_column`). At this point, any temporary columns and triggers created on the affected tables in the `public` schema will also be cleaned up, leaving the table schema in its final state.
```

andrew-farries (changes requested):
Thanks for adding the extra explanations to the docs 🎉 

I've made some suggested changes to better explain the difference between the versioned schema containing the views and the base schema containing the tables.

dolezvo1:
@andrew-farries Thank you, for your notes, I actually wasn't even completely aware of this. I will apply your changes and try again.

andrew-farries (approved):

dolezvo1:
Thank you!

Issue #368, opened 2024-07-09
Incorrect success message if filename and migration name don't match


divyenduz:
``` bash
$ cat sql/1.json
{
  "name": "01_create_users_table",
  "operations": [
    {
      "create_table": {
        "name": "users",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "name",
            "type": "varchar(255)",
            "unique": true
          },
          {
            "name": "description",
            "type": "text",
            "nullable": true
          }
        ]
      }
    }
  ]
}
```

```bash
pgroll start sql/1.json --postgres-url 'postgresql://postgres:postgres@127.0.0.1/cal?sslmode=disable'
 SUCCESS  New version of the schema available under the postgres "public_1" schema
```

The success message incorrectly print schema name based on the filename (`public_1`). But the real schema in the database is called `public_01_create_users_table`

ryanslade:
@divyenduz I just checked and this appears to be fixed, are you happy to close?

divyenduz:
Thanks @ryanslade, just tested it again, it is indeed fixed, closing.

PR #384, opened 2024-08-24
Chore: Add migration start message with schema name

t-ham752:
Hi, I noticed an issue that seemed like a quick fix, so I went ahead and made the changes.
The related issue is here: 
- https://github.com/xataio/pgroll/issues/368

In this PR, I made changes to prioritize using the `name` property in migration file. If `name` is empty, we fall back to using the filename.

I'd appreciate it if you could take a look when you have some time.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -32,7 +32,7 @@ func startCmd() *cobra.Command {
 			}
 			defer m.Close()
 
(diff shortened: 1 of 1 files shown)


exekias (approved):
Great idea, thank you so much for contributing!

PR #392, opened 2024-10-10
Add missing doc comments to exported symbols

inelpandzic:
While going through pgroll codebase I see that it is also really nicely documented overall.

I took this chance to add doc comments to the remaining non-documented exported symbols.


andrew-farries (approved):
Thank you 👍

PR #393, opened 2024-10-11
Fix crash when connecting to PostgreSQL rc version

### LINKED ISSUE #386: pgroll crashes when version includes non-integer characters
Error: unable to retrieve postgres version: sql: Scan error on column index 0, name "split_part": converting driver.Value type string ("17rc1") to a int: invalid syntax

songlim327:
Fixes #386


andrew-farries (approved):
Thank you 👍

ryanslade:
Maybe we can move the version extraction into code to more easily add test cases? Doing a simple `select version()` and then parsing in code may be easier.

PR #396, opened 2024-10-14
Remove direct dependency on golang.org/x/exp

ryanslade:
Most of what we need is already in the stdlib.


andrew-farries (commented):
Would it be worth updating to Go 1.23 instead so that we can use the built-in iterators?

ryanslade:
Sure, we can do that but it forces anyone using pgroll as a library to upgrade too and may force us to tag a new version.

andrew-farries:
I think we should upgrade. We should tag a 0.7 release soon anyway.

ryanslade:
We'll upgrade to Go 1.23 instead and use the new sodlib packages.

PR #398, opened 2024-10-15
Update `docker-compose.yml`

andrew-farries:
Update `docker-compose.yml`

* Remove redundant `version` field.
* Change the port number
* Change the image version to Postgres 17.
* Use an anonymous volume so that it can be cleaned up with `docker compose rm -v`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,15 +1,11 @@
-version: '3.8'
 services:
   db:
-    image: postgres:14-alpine
+    image: 'postgres:17'
     restart: always
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_PASSWORD=postgres
     ports:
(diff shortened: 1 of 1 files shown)


exekias (approved):
will te change in port impact docs? I guess it would require to manually change some URL?

andrew-farries:
The docker compose setup isn't discussed in the docs, so no need to change anything.

exekias:
I guess it still means we deviate from defaults? now setting a URL will be required for init and so, what's the rationale behind the port change?

andrew-farries:
Sorry, didn't see your comment before I merged.

https://github.com/xataio/pgroll/pull/402 changes the port back to `5432`. The motivation was to avoid colliding with other existing Postgres instances that might be running, but changing the port here isn't the best way to solve that.

PR #403, opened 2024-10-15
Speed up `make examples`

ryanslade:
The previous versionw was building the pgroll command before every example run. Instead, build the command once at the beginning and remove it at the end.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,12 +18,14 @@ lint:
 	golangci-lint --config=.golangci.yml run
 
 examples:
-	@go run . init
+	@go build
+	@./pgroll init
 	@for file in examples/*.json; do \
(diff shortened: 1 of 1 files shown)


ryanslade:
@andrew-farries Do we want to somehow use this during CI to confirm examples run correctly?

kvch (approved):

andrew-farries (approved):
Nice 👍 

We already test examples in CI here:
https://github.com/xataio/pgroll/blob/main/.github/workflows/build.yml#L127

It doesn't use `make examples` though; it should probably do so.

ryanslade:
I'll see if I can get it to use the Makefile in this PR

PR #404, opened 2024-10-15
Add support for creating partial indexes on tables

### LINKED ISSUE #322: Support for partial indexes
Our team tried to use pgroll for our new project and found out that it doesn't seem to have a partial index support. 

It seems there were mentions while adding support for indexes (#49) but seems like it is not still implemented.

Since our new project highly depends on partial index, we're wondering if there is any plan for adding this feature.

kvch:
This PR adds a new option to `create_index` operations called `predicate`. When `predicate` is the conditional expression of the partial index. When it is set, pgroll creates a partial index on the table.

Example:
```json
{
  "name": "37_create_partial_index",
  "operations": [
    {
      "create_index": {
        "name": "idx_fruits_id_gt_10",
        "table": "fruits",
        "columns": [
          "id"
        ],
        "predicate": "id > 10"
      }
    }
  ]
}
```

Closes #322

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1043,6 +1043,8 @@ A drop column operation drops a column from an existing table.
 }
 ```
 
+The `down` field above is required in order to backfill the previous version of the schema during an active migration. For instance, in our [example](../examples/09_drop_column.json), you can see that if a new row is inserted against the new schema without a `price` column, the old schema `price` column will be set to `0`.
+
 Example **drop column** migrations:
 
 * [09_drop_column.json](../examples/09_drop_column.json)
--- a/docs/README.md
+++ b/docs/README.md
@@ -971,7 +971,7 @@ A create index operation creates a new btree index on a set of columns.
 Example **create index** migrations:
 
 * [10_create_index.json](../examples/10_create_index.json)
-* [37_create_partial_index.json](../examples/37_create_partal_index.json)
+* [37_create_partial_index.json](../examples/37_create_partial_index.json)
 
 ### Create table
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -963,6 +963,7 @@ A create index operation creates a new btree index on a set of columns.
     "table": "name of table on which to define the index",
     "name": "index name",
     "columns": [ "names of columns on which to define the index" ]
+    "predicate": "conditional expression for defining a partial index",
   }
 }
 ```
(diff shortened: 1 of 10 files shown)


docs/README.md
```diff
@@ -963,13 +963,15 @@ A create index operation creates a new btree index on a set of columns.
...
     "columns": [ "names of columns on which to define the index" ]
+    "predicate": "conditional expression for defining a partial index",
   }
 }
 ```
 
 Example **create index** migrations:
 
 * [10_create_index.json](../examples/10_create_index.json)
+* [37_create_partial_index.json](../examples/37_create_partal_index.json)
```
ryanslade, line 974:
```suggestion
* [37_create_partial_index.json](../examples/37_create_partial_index.json)
```

schema.json
```diff
@@ -223,6 +223,10 @@
         "table": {
           "description": "Name of table on which to define the index",
           "type": "string"
+        },
+        "predicate": {
```
exekias, line 227: I'm wondering if this would be better called `where`, to avoid increasing cognitive load on the user

kvch: I am open to changing it to anything that makes it clearer to the user.

andrew-farries (commented):
This looks good 👍 

We should also update the `read_schema` SQL function to return information about the predicate of a parital index. Right now the new index from the example looks like this:

```json
{
  "indexes": {
    "idx_fruits_id_gt_10": {
      "name": "idx_fruits_id_gt_10",
      "unique": false,
      "columns": [
        "id"
      ]
    }
  }
}

```

(this is part of the `resulting_schema` field after the migration to add the partial index has been applied).

schema.json
exekias: ah just saw this is actual postgres nomenclature in the docs, so we are good

andrew-farries: I think 'predicate' is OK as that is the term used in the Postgres [docs on partial indexes](https://www.postgresql.org/docs/current/indexes-partial.html).

I'm fine with either `where` or `predicate` personally.

exekias: 👍 let's use `predicate`

kvch:
Updated the PR so the output of `analyze` contains the predicate as well:
```json
"idx_fruits_id_gt_10": {
  "name": "idx_fruits_id_gt_10",
  "unique": false,
  "columns": [
    "id"
  ],
  "predicate": "(id \u003e 10)"
}
```

andrew-farries (approved):

Issue #257, opened 2024-01-30
Purpose of drop_column down is unclear in the documentation


simonw:
This section here:

https://github.com/xataio/pgroll/blob/9d2515963865ba4b5b71fe0ff9ea85e8ce9e2ddf/docs/README.md#drop-column

> A drop column operation drops a column from an existing table.
>
> **drop column** operations have this structure:
>
> ```json
> {
>    "drop_column": {
>     "table": "name of table",
>     "column": "name of column to drop",
>     "down": "SQL expression"
>   }
> }

It's not obvious to me what the `"down"` key here is used for.

exekias:
Thank you for opening this issue!

Down migrations are applied to adapt data from the latest schema version into the previous one. During the active migration period, you will have a new schema version that doesn't have the column, while the old schema (that has it) is still available.

If the deployed version of your application expects a value in that column (or there is a constraint requiring it, ie NOT NULL), `down` function can be used to ensure that all inserts/updates happening in the new version of the schema get a valid value backfilled into this column for the old version of the schema (and its clients) to consume.

I realize docs may need clarification on this, we can use this issue to ensure we update them!

PR #405, opened 2024-10-16
Clarify usage of down field when dropping a column.

ryanslade:
Closes https://github.com/xataio/pgroll/issues/257

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1041,7 +1041,7 @@ A drop column operation drops a column from an existing table.
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1041,6 +1041,8 @@ A drop column operation drops a column from an existing table.
 }
 ```
 
+The `down` field above is required in order to backfill the previous version of the schema during an active migration. For instance, in our [example](../examples/09_drop_column.json), you can see that if a new row is inserted against thr new schema without a `price` column, the old schema `price` column will be set to `0`.
```
andrew-farries, line 1044:
```suggestion
The `down` field above is required in order to backfill the previous version of the schema during an active migration. For instance, in our [example](../examples/09_drop_column.json), you can see that if a new row is inserted against the new schema without a `price` column, the old schema `price` column will be set to `0`.
```

andrew-farries (approved):

PR #407, opened 2024-10-16
Remove some lint deprecation warnings

ryanslade:
And switch to use the new `copyloopvar` check for Go > 1.22

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,7 +10,7 @@ linters:
   # https://golangci-lint.run/usage/linters/#enabled-by-default-linters
   enable:
     - errorlint
-    - exportloopref
+    - copyloopvar
     - forcetypeassert
     - goconst
     - gocritic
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -56,16 +56,13 @@ linters-settings:
       - commentFormatting
 
   gofumpt:
-    lang-version: *goVersion
```
exekias, line 59: is this effectively changing the checks? would it be better to update `goVersion` up in this file?

ryanslade: Yeah, that's something I forgot to update when upgrading to Go 1.23. Just changed it, let's see if all the lints still pass.

ryanslade: @exekias All looks good now

exekias: do we need to remove this line now then?

ryanslade: I removed those lines because they were being flagged as deprecated by `golangci-lint`

exekias (approved):

PR #413, opened 2024-10-17
Add support for composite keys in `create table` statements

kvch:
This PR adds support for setting a composite key for a table.
From now on it is possible to set `pk` to `true` in multiple columns in `create_table`.

The create table statement is translated to the following format:

```sql
CREATE TABLE my_table (
    id SERIAL,
    code VARCHAR(255),
    count INTEGER,
    PRIMARY KEY (id, code)
);
```


pkg/migrations/op_create_table_test.go
```diff
@@ -76,6 +76,72 @@ func TestCreateTable(t *testing.T) {
...
+				MustInsert(t, db, schema, "01_create_table", "users", map[string]string{
+					"rand": "123",
+					"name": "Alice",
+				})
+
+				// Data can be retrieved from the new view.
+				rows := MustSelect(t, db, schema, "01_create_table", "users")
+				assert.Equal(t, []map[string]any{
+					{"id": 1, "rand": "123", "name": "Alice"},
+				}, rows)
```
andrew-farries, line 142: I think what we want to assert in this test is that the columns that were marked as a`pk: true` in the migration really are part of the table's primary key.

We can get this information from the system catalogs and have a new assertion like `ColumnMustBePK` in `op_common_test.go`.

Alternatively, the tests could try to insert a duplicate value and assert that the insertion fails with a unique constraint error using the `MustNotInsert` assertion.

kvch: Updated test.

andrew-farries (approved):

Issue #378, opened 2024-07-29
Allow creating different types of indexes and/or return more information about an index type


eemmiillyy:
This is a feature request:

- It would be useful to allow creating indexes other than btree via pgroll.
- It would also be useful to have information created about index types created (also via SQL) on the schema so that we can reconstruct the create index statement for the user that was run.

kvch:
What do you want to do with the reconstructed command? I can return the complete index definition statement. However, with pgroll we create the index concurrently (to avoid locking tables) and the reconstructed statement does not contain it by default.

eemmiillyy:
I want to be able to list the indexes.  I also want to be able to show example SQL for creating indexes. 

Probably returning just the index definition is enough and we don't need to reconstruct any statements.

Does that make sense? @kvch

kvch:
The index definition we can get from PostgreSQL is a complete statement for creating the index e.g. `CREATE INDEX idx_users_name_hash ON public.users USING hash (name) WITH (fillfactor='70')`. But it seems you also need index type and other info for pretty displays.

PR #417, opened 2024-10-18
Support creating different index types

kvch:
This PR adds new options to `create_index` operation:
* `method`: enum for setting different index types. Default: `btree`. Available options: `btree`, `hash`, `gist`, `spgist`, `gin`, `brin`.
* `storage_parameters`: optional storage parameters for the index.

Example operation to create a new hash index with custom fillfactor:

```json
{
  "name": "10_create_index",
  "operations": [
    {
      "create_index": {
        "name": "idx_fruits_name",
        "table": "fruits",
        "columns": [
          "name"
        ],
	"method": "hash",
	"storage_parameters": "fillfactor = 70"
      }
    }
  ]
}
```

Furthermore, `analyze` now return the index method under the name `method` and the definition under `definition`.
Please note that the definition `pgroll` returns is not the SQL statement `pgroll` runs when running `create_index` migrations.

```json
{
 "idx_fruits_name": {
   "name": "idx_fruits_name",
   "unique": false,
   "columns": [
     "name"
   ],
   "method": "hash",
   "definition": "CREATE INDEX idx_fruits_name ON public.fruits USING hash (name) WITH (fillfactor='70')"
 }
}
```

Closes https://github.com/xataio/pgroll/issues/378


pkg/migrations/op_common_test.go
```diff
@@ -296,6 +296,27 @@ func indexExists(t *testing.T, db *sql.DB, schema, table, index string) bool {
 	return exists
 }
 
+func CheckIndexDefinition(t *testing.T, db *sql.DB, schema, table, index, expectedDefinition string) {
```
andrew-farries, line 299: We sould use this also in the `create partial index` test added in a previous PR that currently just checks for index existence.

andrew-farries (commented):
This looks good 👍 

* We should update the documentation with the new index options.
* Consider adding an example migration that shows the new index options and link to it from the docs.

andrew-farries (approved):

PR #426, opened 2024-10-22
Support multiple PKs during backfilling

kvch:
This PR adds support for using multiple primary keys during the backfilling
part of the migration.

In this PR I extracted the statement builder into a separate struct to
make it easier to check for the expected statements.


pkg/migrations/backfill.go
```diff
@@ -84,21 +83,26 @@ func getIdentityColumn(table *schema.Table) *schema.Column {
...
+		lastValues:       nil,
+	}
 }
 
 func (b *batcher) updateBatch(ctx context.Context, conn db.DB) error {
 	return conn.WithRetryableTransaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
 		// Build the query to update the next batch of rows
-		query := b.buildQuery()
+		query := b.statementBuilder.buildQuery(b.lastValues)
+		fmt.Println(query)
```
andrew-farries, line 101: This line should be removed

pkg/migrations/backfill.go
```diff
@@ -107,23 +111,95 @@ func (b *batcher) updateBatch(ctx context.Context, conn db.DB) error {
...
-    ), update AS (
-      UPDATE %[2]s SET %[1]s=%[2]s.%[1]s FROM batch WHERE %[2]s.%[1]s = batch.%[1]s RETURNING %[2]s.%[1]s
-    )
-    SELECT LAST_VALUE(%[1]s) OVER() FROM update
-    `,
-		pq.QuoteIdentifier(b.identityColumn.Name),
-		pq.QuoteIdentifier(b.table.Name),
-		b.batchSize,
-		whereClause)
+	return fmt.Sprintf("SELECT %[1]s FROM %[2]s %[3]sORDER BY %[1]s LIMIT %[4]d FOR NO KEY UPDATE",
```
andrew-farries, line 174:
```suggestion
	return fmt.Sprintf("SELECT %[1]s FROM %[2]s %[3]s ORDER BY %[1]s LIMIT %[4]d FOR NO KEY UPDATE",
```

andrew-farries (commented):
Looks good; just a couple of small things.

pkg/migrations/backfill.go
kvch: For the record, this change adds an extra whitespace before `ORDER BY` if the where clause is missing.

kvch: For the record this change adds an extra whitespace when `where` cluase is missing.

andrew-farries (approved):

Issue #105, opened 2023-09-18
Add knowlege of constraints to internal schema representation


andrew-farries:
Constraints (initially `CHECK` and `FOREIGN KEY`) should be represented in `pg-roll`'s internal schema representation.

Also, `UNIQUE` constraints should be part of the internal schema representation too.

Having constraints represented in the internal schema would enable:
* Better validation of operations involving constraints:
  * Ensure that a constraint of the same name does not exist when adding a new constraint.
  * Ensure that the constraint exists when dropping a constraint. 
* Not having to specify the name of the column on which the constraint is defined when dropping a constraint (and the confusion that occurs when the wrong column is specified)

```[tasklist]
### Tasks
- [x] Validate constraint exists before dropping (https://github.com/xataio/pgroll/pull/252)
- [x] Ensure constraint of same name doesn't exist when adding an `FK` constraint
- [x] Ensure constraint of same name doesn't exist when adding a `CHECK` constraint
- [x] Ensure constraint of same name doesn't exist when adding a `UNIQUE` constraint
- [x] BREAKING CHANGE: no longer require `Column` field when dropping a constraint
```

PR #427, opened 2024-10-22
Validate that unique constraint name is unique

ryanslade:
I'm not sure this is exactly what we want, but putting it up here for review.

This will now catch the case when a unique constraint is added with a name that already exists on the table.
All this really does in shift the check to the `Start` operation instead of the `Complete` operation.

Before this changed we get this error:

```
Failed to complete migration: unable to execute complete operation: pq: index "reviews_username_key" is already associated with a constraint
```

Afterwards we get this:

```
Failed to start migration: migration is invalid: constraint "reviews_username_key" on table "reviews" already exists
```

Part of https://github.com/xataio/pgroll/issues/105

### DIFF
--- a/pkg/migrations/op_common_test.go
+++ b/pkg/migrations/op_common_test.go
@@ -11,9 +11,9 @@ import (
 	"slices"
 	"testing"
 
-	"github.com/xataio/pgroll/internal/testutils"
-
 	"github.com/lib/pq"
+
+	"github.com/xataio/pgroll/internal/testutils"
 	"github.com/xataio/pgroll/pkg/migrations"
 	"github.com/xataio/pgroll/pkg/roll"
 )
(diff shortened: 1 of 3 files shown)


exekias:
good catch! I guess we want the same for the other constraint types? checks & foreign keys

ryanslade:
Yep, those are mentioned in [this issue](https://github.com/xataio/pgroll/issues/105) too. I just wanted to make sure this was the correct approach before doing the others.

ryanslade:
On the one hand, this is nice because we are catching issues early. On the other hand, it means we need to manually think of all the cases where things could go wrong and check for them during validation. I wonder if there is maybe a way to let Postgres find these issues early by running our complete step in an isolated transaction with no intention of committing it first?

ryanslade:
I should have been a little clearer, I didn't mean for just the constraint checks. I'm sure there are a range of things that could go wrong at completion time that we might be able to have Postgres check for us. Something to keep in mind, but agree for this specific issue we can do it manually.

exekias (approved):

exekias:
ah yes, definitely we should try to detect any possible issue at validate or start time. Ideally, a successful start should guarantee complete! so that's something to take into account

PR #429, opened 2024-10-23
Validate check name is unique

ryanslade:
Validate that check name is unique before starting the migration.

### DIFF
--- a/pkg/state/state.go
+++ b/pkg/state/state.go
@@ -542,13 +542,13 @@ func (s *State) Start(ctx context.Context, schemaname string, migration *migrati
 		return nil, err
 	}
 
-	var schema schema.Schema
-	err = json.Unmarshal([]byte(rawSchema), &schema)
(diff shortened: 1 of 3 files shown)


pkg/state/state.go
```diff
@@ -542,13 +542,13 @@ func (s *State) Start(ctx context.Context, schemaname string, migration *migrati
 		return nil, err
 	}
 
-	var schema schema.Schema
-	err = json.Unmarshal([]byte(rawSchema), &schema)
+	var unmarshalledSchema schema.Schema
```
exekias, line 545: this change doesn't seem related to the pr?

exekias (approved):

pkg/state/state.go
ryanslade: Just a small rename to stop a warning about shadowing a package name. I felt it was too small to include in a separate PR.

PR #430, opened 2024-10-23
Defer close of migration file

ryanslade:
More idiomatic to perform file operations in a separate function and defer the closer operation.


ryanslade:
Bit of a pedantic change but I couldn't help myself.

exekias (approved):

PR #431, opened 2024-10-24
Remove Column field from DropConstraint

ryanslade:
We can look this up from the schema so can remove it from the spec.

This is a breaking change since we remove a field from the schema.

Part of https://github.com/xataio/pgroll/issues/105


pkg/migrations/op_drop_constraint.go
```diff
@@ -96,6 +103,18 @@ func (o *OpDropConstraint) Complete(ctx context.Context, conn db.DB, tr SQLTrans
...
+	//
+	// 1. Update the signature of this method to accept a schema object like the others. This would force
+	// us to change our Operation interface.
+	//
+	// 2. Fetch the schema manually from the db using a query and unmarshal it. This feel like a bit of a
+	// specific hack but would save us from changing the Operation interface.
+	//
+	// 3. An even bigger hack, embed the schema in the context for this case only. I don't like it, but
+	// wanted to mention it as an option.
+
```
ryanslade, line 117: @exekias @andrew-farries WDYT? 

Maybe changing the interface to include the schema is the best option (Option 1). Just want some confirmation since maybe there was a reason for omitting it up to this point.

ryanslade: I went with updating the interface as it felt the most consistent with the rest of the code.

kvch (approved):

andrew-farries (approved):
The changes here are fine.

JSON validation rejects unknown fields so existing `drop_constraint` operations that include the `column` field will no longer be valid.

This will require existing migrations to be edited otherwise they won't apply.

ryanslade:
Do you mean existing migrations that exist within xata?

andrew-farries:
> Do you mean existing migrations that exist within xata?

Yes, any existing branches that have a drop constraint migration in their history will fail to have their migations replayed when a user creates a new branch.

PR #434, opened 2024-10-25
Rs/faster tests

ryanslade:


ryanslade:
Simple changes don't seem to make much difference. Closing.

PR #435, opened 2024-10-25
Extract initialisation SQL into separate file

ryanslade:
This makes it easier to edit normally.

I think we may want to pick a SQL formatter for this which we run automatically too.

It looks like prettier already has a SQL formatter so maybe we can just use that since we already use it to format the json schema.


pkg/state/init.sql
```diff
@@ -0,0 +1,340 @@
+CREATE SCHEMA IF NOT EXISTS placeholder;
+
+CREATE TABLE IF NOT EXISTS placeholder.migrations
```
ryanslade, line 3: `placeholder` will be replaced in code.

pkg/state/state.go
```diff
@@ -411,7 +71,8 @@ func (s *State) Init(ctx context.Context) error {
 	}
 
 	// Perform pgroll state initialization
-	_, err = tx.ExecContext(ctx, fmt.Sprintf(sqlInit, pq.QuoteIdentifier(s.schema), pq.QuoteLiteral(s.schema)))
```
ryanslade, line 414: We weren't using the output of `pq.QuoteLiteral`

andrew-farries (approved):
This looks good 👍 

I think we should also find a solution for SQL formatting, either as part of this PR or a follow up.

ryanslade:
> I think we should also find a solution for SQL formatting, either as part of this PR or a follow up.

I'll see if I can get it to use Prettier

ryanslade:
@andrew-farries PTAL

andrew-farries (approved):

kvch (approved):

PR #437, opened 2024-10-28
Respect context cancellation when backing off

ryanslade:


andrew-farries (commented):
Can we add a test here, to ensure that that the backfill is aborted on context cancellation when waiting for a lock?

We already have `TestWithRetryableTransaction` for testing retryable transactions; the new test could be similar to that one.

ryanslade:
@andrew-farries PTAL

andrew-farries (approved):

PR #439, opened 2024-10-28
Validate maximum identifier length

### LINKED ISSUE #64: Add guards for names max length

ryanslade:
It should not exceed the maximum of 63.

Closes #64


kvch (approved):

ryanslade:
(I'm adding tests to all the places I added validation)

ryanslade:
Done, PTAL

PR #442, opened 2024-10-29
Add support for creating unique indexes

kvch:
This PR adds a new attribute to `create_index` operation.
From now on it is possible to create a unique index on one or more
columns in a table.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -973,12 +973,14 @@ A create index operation creates a new index on a set of columns.
 
 The field `method` can be `btree`, `hash`, `gist`, `spgist`, `gin`, `brin`.
 You can also specify storage parameters for the index in `storage_parameters`.
+To create a unique index set `unique` to `true`.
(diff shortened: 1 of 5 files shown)


andrew-farries (approved):
We have a few ways of ways to make unique columns now:

* setting `unique: true` on a column when it's created: [[example](https://github.com/xataio/pgroll/blob/main/examples/01_create_tables.json#L16)]
* using an alter column operation to set a column as unique: [[example](https://github.com/xataio/pgroll/blob/main/examples/15_set_column_unique.json#L8-L10)]

And now creating unique indexes as of this PR.

I think this is fine. One thing to be aware of is that we don't currently duplicate indexes on a column when duplicating it [see https://github.com/xataio/pgroll/issues/227], so if we have a unique index on a column and and then add, for example, a `NOT NULL` constraint to that column, the unique index will be lost.

PR #445, opened 2024-10-31
Add a .ledger file which is built by `make examples`

ryanslade:
Having this file in place makes it harder to commit two examples with the same prefix since it would cause a merge conflict in the ledger.

You can also build only the ledger by running `make ledger`


examples/.ledger
```diff
@@ -0,0 +1,39 @@
+examples/01_create_tables.json
```
ryanslade, line 1: Let me remove the prefix

kvch (approved):

PR #459, opened 2024-11-11
New operation: `create_constraint` and support unique constraints with multiple columns

kvch:
This PR adds a new operation named `create_constraint`. Previously, we only supported adding constraints to columns. This operation lets us define table level constraints including multiple columns.

The operation supports `unique` constraints for now. In a follow-up PR I am adding `foreign_key` and `check` constraints as well.

### Unique

```json
{
  "name": "44_add_table_unique_constraint",
  "operations": [
    {
      "create_constraint": {
        "type": "unique",
        "table": "tickets",
        "name": "unique_zip_name",
        "columns": [
          "sellers_name",
          "sellers_zip"
        ],
        "up": {
          "sellers_name": "sellers_name",
          "sellers_zip": "sellers_zip"
        },
        "down": {
          "sellers_name": "sellers_name",
          "sellers_zip": "sellers_zip"
        }
      }
    }
  ]
}
```

### Related

Created from https://github.com/xataio/pgroll/pull/411


pkg/migrations/op_create_constraint.go
```diff
@@ -0,0 +1,97 @@
...
+			Table: o.Table,
+			Name:  o.Name,
+		}
+		return uniqueOp.Complete(ctx, conn, tr, s)
+	}
+
+	return nil
+}
+
+func (o *OpCreateConstraint) Rollback(ctx context.Context, conn db.DB, tr SQLTransformer, s *schema.Schema) error {
```
kvch, line 42: @andrew-farries I added a rollback operation here to remove the unique index. However, in `op_set_unique.go` the function `Rollback` is a no-op. Which one is the correct rollback process?

andrew-farries: `OpSetUnique.Rollback` is a no-op because the operation is a sub-operation of `OpAlterColumn`. `OpAlterColumn` handles the rollback logic for its sub-operations so there is nothing left for `OpSetUnique.Rollback` to do.

`OpAlterColumn.Rollback` performs a rollback by removing the column that was duplicated and modified along with the triggers that were added to the table.

`OpAlterColumn` contains a lot of the common logic for its various sub-operations. But as `OpCreateConstraint` is a 'standalone' operation we need to implement rollbacks here ourselves.

andrew-farries (commented):
The implementation here adds the `UNIQUE` constraint to the affected columns directly.  In the case of single-column `UNIQUE` constraints added via the `alter_column` operation, the affected column is duplicated and backfilled and the constraint added to the new column. This gives us a 'zero downtime, multiversion'  migration as the old column without the constraint is referenced by the old versioned view and the new column with the constraint is referenced from the new versioned view.

By adding the `UNIQUE` constraint directly to the columns, we break this as now both old and new versions of the schema will be affected by the constraint.

We will need to do the same thing for these potentially multi-column `UNIQUE` constraints as we do for the single column versions:
* on `Start` : duplicate and backfill the affected columns and create a `UNIQUE` index on the new columns.
* on `Complete`: Convert the unique index to a unique constraint.
* on `Rollback`: remove the duplicated columns

The best example of how this works is `op_alter_column` and its various sub-operations.

We will need to implement a few new things:
* Duplicating multiple columns - we need a `_pgroll_new_foo` column for each column covered by the unique index. Currently, we only ever duplicate at most one column per operation.
* Backfill multiple columns - we'll need to run backfills for each column covered by the unique index. Currently, we only backfill at most one column per operation. We either do this by installing multiple triggers or extending the single trigger to handle filling more than one field.

After `Start` we need the schema to look like this:

```
| ticket_id                | integer                |  not null default nextval('_pgroll_new_tickets_ticket_id_seq'::regclass) | plain    | <null>       | <null>      |
| sellers_name             | character varying(255) |  not null                                                                | extended | <null>       | <null>      |
| sellers_zip              | integer                |  not null                                                                | plain    | <null>       | <null>      |
| _pgroll_new_sellers_name | character varying(255) |  not null                                                                | extended | <null>       | <null>      |
| _pgroll_new_sellers_zip  | integer                |  not null                                                                | plain    | <null>       | <null>      |

Indexes:
    "_pgroll_new_tickets_pkey" PRIMARY KEY, btree (ticket_id)
    "unique_zip_name" UNIQUE, btree (_pgroll_new_sellers_name, _pgroll_new_sellers_zip)
Triggers:
<up and down triggers to handle the backfill and dual write>
```

kvch:
@andrew-farries I have updated the PR with backfilling. The example migration file has changed, let me know what you think of this format. (I am not really happy with it, but this was my best idea :D)

pkg/migrations/op_create_constraint_test.go
```diff
@@ -0,0 +1,163 @@
...
+				// The index has been created on the underlying table.
+				IndexMustExist(t, db, schema, "users", "unique_name")
+			},
+			afterRollback: func(t *testing.T, db *sql.DB, schema string) {
+				// The index has been dropped from the the underlying table.
+				IndexMustNotExist(t, db, schema, "users", "uniue_name")
+			},
+			afterComplete: func(t *testing.T, db *sql.DB, schema string) {
+				// Complete is a no-op.
+			},
```
andrew-farries, line 63: I think these tests should be extended to actually check that the unique constraint is effective on the `users` column in the new schema version but not on the old.

There is an example of how this can look in `op_set_unique_test.go`.

andrew-farries (commented):
This is looking good. Left some comments.

docs/README.md
```diff
@@ -1037,6 +1039,32 @@ Example **create table** migrations:
...
+**create constraint** operations have this structure:
+
+```json
+{
+  "create_constraint": {
+    "table": "name of table",
+    "name": "my_unique_constraint",
+    "columns": ["list of columns"],
+    "type": "unique"
+  }
```
andrew-farries, line 1059: I think `up` and `down` should be mentioned here too.

schema.json
```diff
@@ -432,6 +432,42 @@
...
+        },
+        "type": {
+          "description": "Type of the constraint",
+          "type": "string",
+          "enum": ["unique"]
+        },
+        "up": {
+          "type": "object",
+          "description": "SQL expression of up migration by column"
+        },
```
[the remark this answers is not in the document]
andrew-farries: This may not be possible; in any case we decided not to implement this here.

pkg/migrations/op_create_constraint.go
```diff
@@ -0,0 +1,216 @@
...
+	_, err := conn.ExecContext(ctx, fmt.Sprintf("CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS %s ON %s (%s)",
+		pq.QuoteIdentifier(o.Name),
+		pq.QuoteIdentifier(o.Table),
+		strings.Join(quotedTemporaryNames(o.Columns), ", "),
+	))
+
+	return err
+}
+
+func quotedTemporaryNames(columns []string) []string {
```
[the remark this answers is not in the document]
kvch: nice catch

andrew-farries (approved):

PR #463, opened 2024-11-13
Add a `pgroll pull` subcommand

andrew-farries:
Add a new `pgroll pull` command that pulls the complete migration history for a schema from the `migrations` table in the target database and dumps the migrations to disk.

## Example

Given the migration history in `pgroll.migrations` after applying all `example/` migrations in this repo:

```
+--------+-----------------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
| schema | name                                    | migration                                                                                                                                                                                                                                              >
|--------+-----------------------------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
| public | 01_create_tables                        | {"name": "01_create_tables", "operations": [{"create_table": {"name": "customers", "columns": [{"pk": true, "name": "id", "type": "integer"}, {"name": "name", "type": "varchar(255)", "unique": true}, {"name": "credit_card", "type": "text", "nullab>
| public | 02_create_another_table                 | {"name": "02_create_another_table", "operations": [{"create_table": {"name": "products", "columns": [{"pk": true, "name": "id", "type": "serial"}, {"name": "name", "type": "varchar(255)", "unique": true}, {"name": "price", "type": "decimal(10,2)"}>
| public | 03_add_column_to_products               | {"name": "03_add_column_to_products", "operations": [{"add_column": {"up": "UPPER(name)", "table": "products", "column": {"name": "description", "type": "varchar(255)", "nullable": true}}}, {"add_column": {"table": "products", "column": {"name": ">
| public | 04_rename_table                         | {"name": "04_rename_table", "operations": [{"rename_table": {"to": "clients", "from": "customers"}}]}                                                                                                                                                  >
| public | 05_sql                                  | {"name": "05_sql", "operations": [{"sql": {"up": "CREATE TABLE users (id INTEGER PRIMARY KEY, name TEXT)", "down": "DROP TABLE users"}}]}                                                                                                              >
| public | 06_add_column_to_sql_table              | {"name": "06_add_column_to_sql_table", "operations": [{"add_column": {"up": "UPPER(name)", "table": "users", "column": {"name": "description", "type": "varchar(255)", "nullable": true}}}]}                                                           >
...
```

Run:

```bash
$ pgroll pull migrations/
```

This will pull all migrations from the `pgroll.migrations` table into the `migrations/` directory:

```
$ ls migrations/

01_create_tables.json
02_create_another_table.json
03_add_column_to_products.json
04_rename_table.json
05_sql.json
06_add_column_to_sql_table.json
...
```

The optional `--with-prefixes` flag prefixes each migration name with its position in the schema history:

```bash
$ pgroll pull --with-prefixes migrations/
```

This produces the following files:

```
$ ls migrations/

0001_01_create_tables.json
0002_02_create_another_table.json
0003_03_add_column_to_products.json
0004_04_rename_table.json
0005_05_sql.json
0006_06_add_column_to_sql_table.json
...
```

The `--with-prefixes` flag ensures that files are sorted lexicographically by their time of application.

### DIFF
--- a/cmd/pull.go
+++ b/cmd/pull.go
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/xataio/pgroll/cmd/flags"
+	"github.com/xataio/pgroll/pkg/state"
+
+	"github.com/spf13/cobra"
+)
+
+func pullCmd() *cobra.Command {
+	opts := map[string]string{
+		"p": "prefix each migration filename with its position in the schema history",
+	}
+	var withPrefixes bool
+
+	pullCmd := &cobra.Command{
+		Use:   "pull <target directory>",
+		Short: "pull migration history from the target database and write it to disk",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			targetDir := args[0]
+
+			state, err := state.New(ctx, flags.PostgresURL(), flags.StateSchema())
+			if err != nil {
+				return err
+			}
+			defer state.Close()
+
+			migs, err := state.SchemaHistory(ctx, flags.Schema())
+			if err != nil {
+				return fmt.Errorf("failed to read schema history: %w", err)
+			}
+
+			for i, mig := range migs {
+				prefix := ""
+				if withPrefixes {
+					prefix = fmt.Sprintf("%04d", i+1) + "_"
+				}
+				err := mig.WriteToFile(targetDir, prefix)
+				if err != nil {
+					return fmt.Errorf("failed to write migration %q: %w", mig.Migration.Name, err)
+				}
+			}
+			return nil
+		},
+	}
+
+	pullCmd.Flags().BoolVarP(&withPrefixes, "with-prefixes", "p", false, opts["p"])
+
+	return pullCmd
+}
--- a/pkg/state/history.go
+++ b/pkg/state/history.go
@@ -1,0 +1,84 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package state
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/migrations"
+)
+
+// Migration represents a single migration in the migration history
+// of a schema
+type Migration struct {
+	Migration migrations.Migration
+	CreatedAt time.Time
+}
+
+// SchemaHistory returns all migrations applied to a schema in ascending
+// timestamp order
+func (s *State) SchemaHistory(ctx context.Context, schema string) ([]Migration, error) {
+	rows, err := s.pgConn.QueryContext(ctx,
+		fmt.Sprintf(`SELECT name, migration, created_at
+			FROM %s.migrations
+			WHERE schema=$1 ORDER BY created_at`,
+			pq.QuoteIdentifier(s.schema)), schema)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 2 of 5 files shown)


kvch (approved):

pkg/state/history.go
```diff
@@ -0,0 +1,86 @@
...
+type Migration struct {
+	Migration migrations.Migration
+	CreatedAt time.Time
+}
+
+// SchemaHistory returns all migrations applied to a schema in ascending
+// timestamp order
+func (s *State) SchemaHistory(ctx context.Context, schema string) ([]Migration, error) {
+	rows, err := s.pgConn.QueryContext(ctx,
+		fmt.Sprintf("SELECT name, migration, created_at "+
```
ryanslade, line 28: nit: You can use backticks to quote the string and include newlines so that you don't need to break it up with `+`

pkg/state/history.go
```diff
@@ -0,0 +1,86 @@
...
+		var name, rawMigration string
+		var createdAt time.Time
+
+		rows.Scan(&name, &rawMigration, &createdAt)
+		if err != nil {
+			return nil, err
+		}
+
+		var mig migrations.Migration
+		json.Unmarshal([]byte(rawMigration), &mig)
```
ryanslade, line 49: You're not assigning to `err` here

pkg/state/history.go
```diff
@@ -0,0 +1,86 @@
...
+		return err
+	}
+
+	// Ensure the target directory exists
+	err = os.MkdirAll(targetDir, 0o755)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(filePath, bytes, 0o600)
```
ryanslade, line 85: nit: You could create a file first and use a json encoder to write directly to it rather than writing to memory and then the file.

exekias (commented):
nice! should we add something to the docs?

pkg/state/history.go
andrew-farries: done

pkg/state/history.go
andrew-farries: good spot, thanks

pkg/state/history.go
andrew-farries: thanks, done

andrew-farries:
> nice! should we add something to the docs?

If we're happy that `pgroll pull` is the right command name, yes.

I've added documentation in the most recent commit on the branch.

ryanslade (approved):

cmd/pull.go
```diff
@@ -0,0 +1,52 @@
...
+				err := mig.WriteToFile(targetDir, withTimestamps)
+				if err != nil {
+					return fmt.Errorf("failed to write migration %q: %w", mig.Migration.Name, err)
+				}
+			}
+			return nil
+		},
+	}
+
+	pullCmd.Flags().BoolVarP(&withTimestamps, "with-timestamps", "t", false, opts["t"])
```
exekias, line 49: I've been thinking about this one, what do you think about relying on numbers here? we could do `0001_` etc, looking at the example output, timestamps are not the most readable thing

exekias (approved):

cmd/pull.go
andrew-farries: I think you're right.

The flag is now `--with-prefixes` and prefixes each filename with `0001_` etc as you suggest.

PR description and docs are updated to reflect the change.

exekias (approved):

exekias:
Awesome! I wonder if we should (not in this PR) update our examples folder to use these prefixes, while taking the numbers out of the migration names

andrew-farries:
> Awesome! I wonder if we should (not in this PR) update our examples folder to use these prefixes, while taking the numbers out of the migration names

Taking migration numbers out of the migration names is a good idea I think 👍

Issue #446, opened 2024-11-01
Difficulties integrating into project


ericwooley:
Hello!

I love the project and the goal.

However, as it stands, it's pretty hard to integrate into a project with multiple branches and developers. 

Presumably, each developer, working on different branches would have different migration files. Something like this

```
migrations
├── create_book.json             # dev 1 on branch book which was branched off main
├── create_checkouts.json.   # dev 2 on branch checkouts which was branched off of book
└── create_library.json          # dev 3 on branch library which was branched off main
```

When they all get merged it's hard to know whats been merged, and in what order to run them. 

i've resorted to scripting the creation of migrations so that they have time stamps, so they run in order of the time the developer created them, which seems to work ok, but then the migration script on pull down needs to figure out which migrations to run, and which ones not to run for this environment, and there doesn't seem to be a way to check that using the CLI. Currently I'm having to resort to querying the migrations table myself to see what needs to be run.

Another option i considered was storing individual operations in a folder, based on release, but this also gets hard to manage as different releases might happen with different features etc.... Which would require moving files around as part of the release process. 

Guidance on this would be helpful, or possibly allowing a status query on individual migrations files, so that I can check if each migration has already been run, and skip it if so.

I'm using this for a typical SASS deployment process for now, so most of the time I will just be running `pgroll start <file> --complete`.  But I like the idea of using this for gnarlier future migrations as well, when necessary.

exekias:
Thank you for opening this one @ericwooley!

We recently discussed this, and this is a possible approach we could take:

* Migration files must be in a well-known (configurable) folder, ie `pgroll`
* Migration files should be sorted by name, ie by leveraging timestamps (as you do), or numbering (`001_`, etc)
* We introduce a `pgroll migrate` subcommand, that will apply all pending migrations to the database, and leave the last one started (without complete). This command will auto-complete any previous migration until it reaches the last one in the `pgroll` folder. You would run this command from your CI to ensure the DB has all the migrations in place, with the last one started, so previous version is still available for previous application instances.
* We introduce a `pgroll latest_version_schema` (better name pending?) to ease the task of configuring clients. This command will output the schema to be used in the `search_path` to access the latest version (using local files in `pgroll` folder as the source of truth).

I believe these 2 new commands would ease local development and continuous deployment.


Please let us know what you think, we are really interested in your feedback!:

ericwooley:
@exekias 

Those sound like great improvements. I believe they would relieve the issues I'm running into. 

[...]

PR #414, opened 2024-10-17
New command: `bootstrap`

kvch:
This PR adds a new command called `bootstrap`. It runs all the migrations in the specified folder. Migration files are JSON files, ordered alphabetically.

Help:
```
> go run . bootstrap -h
Bootstrap a new database from a directory of migration files. All files in the directory will be executed
in alphabetical order. All migrations are completed.

Usage:
  pgroll bootstrap <folder> [flags]

Flags:
  -h, --help   help for bootstrap
```

Closes https://github.com/xataio/pgroll/issues/269

PR #465, opened 2024-11-14
Add `pgroll migrate` subcommand

andrew-farries:
Add a `pgroll migrate` subcommand.

## Documentation
`pgroll migrate` applies all outstanding migrations from a source directory to the target database.

Assuming that migrations up to and including migration `40_create_enum_type` from the [example migrations directory](https://github.com/xataio/pgroll/tree/main/examples) have been applied, running:

```
$ pgroll migrate examples/
```

will apply migrations from `41_add_enum_column` onwards to the target database.

If the `--complete` flag is passed to `pgroll migrate` the final migration to be applied will be completed. Otherwise the final migration will be left active (started but not completed).

## Notes:
* If no migrations have yet been applied to the target database, `migrate`  applies all of the migrations in the source directory.
* This PR removes the `pgroll bootstrap` command (https://github.com/xataio/pgroll/pull/414) as it is equivalent to running `pgroll migrate <directory> --complete` against a fresh database.

Part of https://github.com/xataio/pgroll/issues/446


kvch (approved):

cmd/migrate.go
```diff
@@ -0,0 +1,107 @@
...
+			ctx := cmd.Context()
+			migrationsDir := args[0]
+
+			m, err := NewRoll(ctx)
+			if err != nil {
+				return err
+			}
+			defer m.Close()
+
+			latestVersion, err := m.State().LatestVersion(ctx, m.Schema())
```
exekias, line 33: can we move all this logic into the pgroll library? Ideally we can use this as a lib too

cmd/migrate.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, fmt.Errorf("reading migration directory: %w", err)
+	}
+
+	if latestVersion == nil {
+		return files, nil
+	}
+
+	var idx int
+	for _, file := range files {
+		bytes, err := os.ReadFile(file)
```
ryanslade, line 89: You could marshal directly from the file using a json.Decoder

cmd/migrate.go
andrew-farries: What would the API look like to apply outstanding migrations and where would it live?

something like this?:

```go
m : = roll.New(ctx, pgURL, schema, state, opts)

// Apply outstanding migrations 
err := m.Migrate("./migrations/")
```

or would it be better to have a lower level API that returns unapplied migration files:

```go
m : = roll.New(ctx, pgURL, schema, state, opts)

// get outstanding migration files
err, files := m.UnappliedMigrations("./migrations")
for _, file := range files {
  // start and complete each file
}
```

Do you know how you would like to use this in your existing integration?

exekias: example on the top would work!

Ideally we work on top of the some `io/fs` abstraction, so I can leverage `embed.FS` as a parameter.

cmd/start.go
```diff
@@ -71,9 +73,6 @@ func runMigrationFromFile(ctx context.Context, m *roll.Roll, fileName string, co
 	}
 
 	version := migration.Name
-	if version == "" {
-		version = strings.TrimSuffix(filepath.Base(fileName), filepath.Ext(fileName))
-	}
```
exekias, line 76: 👍

pkg/roll/roll.go
```diff
@@ -162,3 +168,56 @@ func (m *Roll) Close() error {
...
+
+			migration, err := migrations.ReadMigration(file)
+			if err != nil {
+				return nil, fmt.Errorf("reading migration file %q: %w", file, err)
+			}
+
+			idx++
+			if migration.Name == *latestVersion {
+				break
+			}
```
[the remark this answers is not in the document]
andrew-farries: This could occur if a user runs `pgroll migrate foo/` where `foo` is a directory containing migrations other than the set applied to the target database.

This is a specific case of the following behaviour:

If I have a target schema at version `03_foo` and I have a `migrations/` directory like this:

```
01_apples.json
02_bananas.json
03_foo.json
04_cherries.json
```

`pgroll migrate migrations/` will apply `04_cherries.json` even if `01_apples` and `02_bananas` were never applied to the target schema; all migrations from the latest version of the target schema onwards are applied without checking if versions prior to `03_foo` were applied.

For the common case where a user is applying the right migrations directory this behaviour is fine, but we could be more strict and fail if the exact sequence of migrations on disk has not been applied.

WDYT?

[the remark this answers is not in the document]
exekias: yes, I believe that would be my expectation as a user?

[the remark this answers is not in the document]
andrew-farries: This is implemented and tested in 3c5824caa7bb3ac9972e82eddd4172409b5dea78.

If the local migration directory doesn't match the remote history, `Roll.UnappliedMigrations` now returns an `ErrMismatchedMigration`.

exekias (approved):

PR #464, opened 2024-11-14
Add support for creating `CHECK` constraints with `create_constraint`

kvch:
This PR introduces a new constraint `type` to `create_constraint` operation called `check`. Now it is possible to create check constraints on multiple columns.

### Example
```json
{
  "name": "45_add_table_check_constraint",
  "operations": [
    {
      "create_constraint": {
        "type": "check",
        "table": "tickets",
        "name": "check_zip_name",
        "columns": [
          "sellers_name",
          "sellers_zip"
        ],
        "check": "sellers_name ~ 'Alice' AND sellers_zip IS NOT NULL",
        "up": {
          "sellers_name": "Alice",
          "sellers_zip": "(SELECT CASE WHEN sellers_zip IS NOT NULL THEN sellers_zip ELSE '00000' END)"
        },
        "down": {
          "sellers_name": "sellers_name",
          "sellers_zip": "sellers_zip"
        }
      }
    }
  ]
}
```

PR #466, opened 2024-11-14
Duplicate unique and check constraints correctly

kvch:
Previously, unique and check constraints with multiple columns were not duplicated correctly.

We had two issues:
1. Pgroll tried to create the same duplicated index for each renamed column. From now on pgroll creates the index once with the new column names.
2. Pgroll tried to convert the unique index to a constraint multiple times. From now on, the index is only converted once.

Required by https://github.com/xataio/pgroll/pull/464

### DIFF
--- a/pkg/migrations/duplicate.go
+++ b/pkg/migrations/duplicate.go
@@ -14,12 +14,10 @@ import (
 	"github.com/xataio/pgroll/pkg/schema"
 )
 
-// Duplicator duplicates a column in a table, including all constraints and
+// ColumnDuplicator duplicates a column in a table, including all constraints and
 // comments.
-// Column must be set with WithColumn before calling Duplicate.
-type Duplicator struct {
-	conn              db.DB
-	table             *schema.Table
+type ColumnDuplicator struct {
+	duplicator        *duplicator
 	column            *schema.Column
 	asName            string
 	withoutNotNull    bool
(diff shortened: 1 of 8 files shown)


kvch:
It is highly likely we need to adjust the check constraint migrations as well. I will address those issues in a follow-up PR.

pkg/migrations/duplicate.go
```diff
@@ -171,16 +182,30 @@ func (d *Duplicator) Duplicate(ctx context.Context) error {
...
 
 		if slices.Contains(uc.Columns, d.column.Name) {
+			cols := uc.Columns
+			// Drop the existing unique index if it is a duplicated unique constraint
+			if existingConstraint, ok := d.duplicatedUnique[DuplicationName(uc.Name)]; ok {
+				dropIndex := fmt.Sprintf("DROP INDEX CONCURRENTLY %s", pq.QuoteIdentifier(DuplicationName(uc.Name)))
+				_, err := d.conn.ExecContext(ctx, dropIndex)
+				if err != nil {
+					return err
+				}
```
andrew-farries, line 192: I'm not sure that creating and then dropping indexes duplicated by previous columns is the way to go here.

Previously the column duplicator was stateless; it duplicated columns without reference to other columns that may have been duplicated previously. The need to duplicate multi-column constraints means it has to become stateful to some extent.

Some options for how to handle column duplication now that it needs to be stateful:

* Introduce some kind of duplication context type that can hold state between calls to the duplicator.
* Redesign the duplicator to be a multi-column duplicator.

Not sure OTOH which is better (maybe the second option?) but I think we need to accept that column duplication is now stateful in the presence of multi-column constraints and design it accordingly.

kvch: I already implemented the multi-column duplicator. However, i decided to submit this as it was a smaller change. Anyway the multi-column duplicator is probably a better approach as i need to adjust the duplication of check constraints as well.

kvch:
I pushed the multi-column duplicator change.

pkg/migrations/duplicate.go
```diff
@@ -187,6 +154,165 @@ func (d *Duplicator) Duplicate(ctx context.Context) error {
...
+		}
+
+		// Duplicate the column's comment
+		if err := cg.duplicator.duplicateComment(ctx, column, asName); err != nil {
+			return err
+		}
+	}
+
+	// Generate SQL to duplicate any check constraints on the column group.
+	for _, cc := range cg.duplicator.table.CheckConstraints {
```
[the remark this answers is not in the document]
kvch: You are right option 2 only works if the column has default or not null constraint, but fails to duplicate if it has unique constraint. I will add a fix and tests.

pkg/migrations/duplicate.go
```diff
@@ -17,174 +17,283 @@ import (
...
+				column:   column,
+				asName:   TemporaryName(column.Name),
+				withType: column.Type,
+			},
+		},
+		withoutConstraint: make([]string, 0),
+	}
+}
+
+func NewColumnGroupDuplicator(conn db.DB, table *schema.Table, columns []*schema.Column) *Duplicator {
```
andrew-farries, line 66: I think the `NewColumnDuplicator` and `NewColumnGroupDuplicator` constructor functions could be combined too.

andrew-farries (approved):
Great 🎉 

Approved with one more suggesion. 👍

kvch:
Updated the PR with the suggestion, and merging once the tests pass.

Issue #408, opened 2024-10-16
Create performance benchmarks for key `pgroll` features


andrew-farries:
Gather benchmarks data for the following parts of `pgroll`:

- [x] **Backfill duration** - How long does it take to perform backfill operations on a table of some fixed size (say 10^7 rows)?
- [x]  **Effect of dual writes** - What overhead do the up/down triggers incur on `UPDATE` heavy tables?
- [x] **read_schema query performance**  - Benchmark the performance of the `read_schema` query, run on every DDL statement to capture 'inferred' migrations.
- [x] Persist results somewhere so we can see changes over time

Having these benchmarks in place would allow us to measure performance improvements over time and avoid regressions.

ryanslade:
I'd like to have a go at this. 

In a perfect world we'd probably want to run these against every commit, but I imagine they may take a while to run and I don't want to affect the velocity of getting things into `main`. Maybe a compromise is that we spin up an environment once a day and run the tests against all new commits?

Apart from actually writing the benchmarks, we need to decide on a few things:

- How often do we run them? I suggest once a day as mentioned above
- Where do we run them? I think we may want to spin up a dedicated environment in EC2 so that the results are consistent
- Where do we store results? Ideally, since this is an open source project, we may want the results to be public. Perhaps we can upload results to a wiki / docs area in this repo?

Anything else?

andrew-farries:
I think what you suggest is a good start. We want the benchmarks for a couple of reasons:

* Guard against performance regressions
* Have benchmarks available as part of the public documentation for the repository.

I suggest running the benchmarks as a separate workflow that is automatically run on changes to `main` and that can also be invoked manually on branches. 

A consistent environment in terms of hardware and probably also software (maybe run the benchmarks in a container) is a must too.

Results could be uploaded to object storage and pulled from there into our docs.

ryanslade:
@andrew-farries @exekias The basic benchmarks are all done now, shall we close this? 

We can add new issues if we want to add more benchmarks, change how they're run etc.

andrew-farries:
Yes, let's close it.

PR #467, opened 2024-11-18
Build static HTML page using the latest benchmark results

ryanslade:
This uses a GitHub action to download benchmark data from S3. It then uses a static generator to create charts per benchmark / Postgres version. These results are then bundled up using Jekyll which publishes the README, available at:

https://xataio.github.io/pgroll/

The benchmarks themselves can be see at https://xataio.github.io/pgroll/benchmarks.html (and are also linked from the README)

In order to make this work and also not require users of pgroll as a library to import the charting library we needed to create a new module in `/dev`. We needed to duplicate the definition of the benchmark results structs since we couldn't import them from the main `pgroll` module without actually publishing a new version of the module. Once this module has been published it will be possible.

Part of https://github.com/xataio/pgroll/issues/408


benchmark-results/build.go
```diff
@@ -0,0 +1,228 @@
+// SPDX-License-Identifier: Apache-2.0
```
andrew-farries, line 1: Should this be under a `cmd/` dir rather than at the top level?

benchmark-results/testdata/benchmark-results.json
```diff
@@ -0,0 +1,5 @@
+{"GitSHA":"9c7821fb64e345e3d7b18aaefe010bb5ffb8dddc","PostgresVersion":"14.8","Timestamp":1731919355,"Reports":[{"Name":"BenchmarkBackfill/10000","RowCount":10000,"Unit":"rows/s","Result":53318.89969391059},{"Name":"BenchmarkBackfill/100000","RowCount":100000,"Unit":"rows/s","Result":20771.251554603292},{"Name":"BenchmarkBackfill/300000","RowCount":300000,"Unit":"rows/s","Result":9355.33788169637},{"Name":"BenchmarkWriteAmplification/NoTrigger/10000","RowCount":10000,"Unit":"rows/s","Result":349447.42577457114},{"Name":"BenchmarkWriteAmplification/NoTrigger/100000","RowCount":100000,"Unit":"rows/s","Result":347618.8045838434},{"Name":"BenchmarkWriteAmplification/NoTrigger/300000","RowCount":300000,"Unit":"rows/s","Result":332593.7044319948},{"Name":"BenchmarkWriteAmplification/WithTrigger/10000","RowCount":10000,"Unit":"rows/s","Result":77669.81724959961},{"Name":"BenchmarkWriteAmplification/WithTrigger/100000","RowCount":100000,"Unit":"rows/s","Result":76891.54517908255},{"Name":"BenchmarkWriteAmplification/WithTrigger/300000","RowCount":300000,"Unit":"rows/s","Result":79160.3927250836}]}
```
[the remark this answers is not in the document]
ryanslade: I'm going to write a small test that uses this data.

I assumed that using env vars for config would be easier, especially from GitHub actions but it actually ended up not. I'll switch it to use command line args.

benchmark-results/build.go
ryanslade: Yeah, I want to move it somewhere else. Ideally I'd like to separate it from pgroll itself so that when users import pgroll as a library they don't also need to pull in the charting library. 

To do this I may stick it in somewhere like maybe `/dev/` so indicate that it's for internal use and give it it's own `go.mod` file. WDYT?

andrew-farries: Could the code to build the benchmarks be a test so that the dependency isn't required outside of test code? That would solve the problem of people having the extra charting dependencies.

Failing that, having it be it's own module could work.

ryanslade: I tried moving it all into test code but when I run `go mod tidy` in the root it still pulls in as a direct dependency. I think this would then cause it to be pulled in when users import pgroll. 

I'll have to create a separate module for it.

ryanslade:
@andrew-farries PTAL, I think this is ready to go now.

andrew-farries (approved):

PR #469, opened 2024-11-19
Add a `pgroll latest` command

andrew-farries:
Add a `pgroll latest` command to show the latest version in either the target database or a local directory of migration files.

## Documentation

`pgroll latest` prints the latest schema version in either the target database or a local directory of migration files. 

By default, `pgroll latest` prints the latest version in the target database. Use the `--local` flag to print the latest version in a local directory of migration files instead.

In both cases, the `--with-schema` flag can be used to prefix the latest version with the schema name.

#### Database

Assuming that the [example migrations](https://github.com/xataio/pgroll/tree/main/examples) have been applied to the `public` schema in the target database, running:

```
$ pgroll latest 
```

will print the latest version in the target database:

```
45_add_table_check_constraint
```

The exact output will vary as the `examples/` directory is updated.

#### Local

Assuming that the [example migrations](https://github.com/xataio/pgroll/tree/main/examples) are on disk in a directory called `examples`, running:

```
$ pgroll latest --local examples/
```

will print the latest migration in the directory:

```
45_add_table_check_constraint
```

The exact output will vary as the `examples/` directory is updated.

---
Part of https://github.com/xataio/pgroll/issues/446

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,6 +13,8 @@
     * [complete](#complete)
     * [rollback](#rollback)
     * [status](#status)
+    * [migrate](#migrate)
+    * [latest](#latest)
     * [pull](#pull)
 * [Operations reference](#operations-reference)
     * [Add column](#add-column)
@@ -537,6 +539,8 @@ The `pgroll` CLI offers the following subcommands:
 * [complete](#complete)
 * [rollback](#rollback)
 * [status](#status)
+* [migrate](#migrate)
+* [latest](#latest)
 * [pull](#pull)
 
 The `pgroll` CLI has the following top-level flags:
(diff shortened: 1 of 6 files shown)


docs/README.md
```diff
@@ -638,6 +642,44 @@ will apply migrations from `41_add_enum_column` onwards to the target database.
...
+$ pgroll latest examples/ --with-schema --schema foo
+```
+
+This prints:
+
+```
+foo_45_add_table_check_constraint
+```
+
+The `pgroll latest` command is useful in scripts that need to retrieve the latest version from a set of migrations files, for example to set the Postgres `search_path` in a CD workflow.
```
kvch, line 679: I wonder if we need to be more explicit. Is this the last migration that was completed? Or is this the last in progress migration?

kvch (approved):

docs/README.md
andrew-farries: It's actually neither. 

The command prints the latest version it finds in the argument directory of migrations. That migration may not have even been applied to the database yet.

The use case is to run this command in a CD workflow where the service to be deployed needs to set the Postgres `search_path` as part of the DB connection string. The idea is that somewhere in the workflow this is run:

```
DB_SCHEMA=$(pgroll latest --with-schema migrations/)
// $DB_SCHEMA now contains something like "public_03_add_constraint"
```

then the connection string for the application can be built using `$DB_SCHEMA`.

Maybe `pgroll latest` is not the best name for this command, but I'm not sure of a better name.

exekias (approved):

docs/README.md
kvch: Yes, it is not really clear. Either way I do not yet have a better name for this. Instead we could document it a bit more verbosely. Could you please add a note about this to the documentation? I assumed based on reading the code that it has nothing to do with the applied migrations. But I would prefer that users avoid this step. :)

ryanslade (approved):

docs/README.md
andrew-farries: I've updated the command so that it can now retrieve the latest schema version from either the target database or a local migrations directory.

I think this addresses the confusion about what the command is operating on.

andrew-farries:
I've updated the command so that it can now retrieve the latest schema version from either the target database or a local migrations directory. The PR description and documentation has been updated with the new behaviour.

I think this addresses the confusion about what the command is operating on that @kvch raised in https://github.com/xataio/pgroll/pull/469#discussion_r1848160792.

PR #472, opened 2024-11-20
Add a benchmark for read_schema

ryanslade:
Add a benchmark for the `read_schema` function which has caused regressions in the past.

Part of https://github.com/xataio/pgroll/issues/408

### DIFF
--- a/internal/benchmarks/benchmarks_test.go
+++ b/internal/benchmarks/benchmarks_test.go
@@ -21,7 +21,10 @@ import (
 	"github.com/xataio/pgroll/pkg/roll"
 )
 
-const unitRowsPerSecond = "rows/s"
+const (
+	unitRowsPerSecond       = "rows/s"
+	unitExecutionsPerSecond = "executions/s"
+)
 
 var (
 	rowCounts = []int{10_000, 100_000, 300_000}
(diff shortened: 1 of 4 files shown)


internal/benchmarks/benchmarks_test.go
```diff
@@ -173,6 +173,7 @@ func BenchmarkReadSchema(b *testing.B) {
 		// We don't want this benchmark to test the network so instead we run the actual function in a tight
 		// loop within a single execution.
 		executions := 10000
+		// nolint:gosec
```
exekias, line 176: instead of ignoring this, I believe the linter wants you to pass the executions number as an arg in `ExecContext`

internal/benchmarks/benchmarks_test.go
```diff
@@ -154,6 +157,38 @@ func BenchmarkWriteAmplification(b *testing.B) {
...
+	ctx := context.Background()
+	testSchema := testutils.TestSchema()
+	var opts []roll.Option
+
+	testutils.WithMigratorInSchemaAndConnectionToContainerWithOptions(b, testSchema, opts, func(mig *roll.Roll, db *sql.DB) {
+		b.Cleanup(func() {
+			require.NoError(b, mig.Close())
+		})
+
+		setupInitialTable(b, ctx, testSchema, mig, db, 1)
```
exekias, line 170: it seems clear that for the shake of this test and the others, we probably want a more complex schema here, right? not for this PR

exekias (approved):

internal/benchmarks/benchmarks_test.go
ryanslade: Yeah, agree. I'm not sure exactly how the performance of `read_schema` would change depending on the complexity of the schema but I guess that's something we could actually benchmark too :)

exekias: ++

PR #477, opened 2024-11-21
Change SQL fomatting tool to `pgformatter`

andrew-farries:
Change the SQL formatting tool used by `make format` from [sql-formatter](https://github.com/sql-formatter-org/sql-formatter) to [pgFormatter](https://github.com/darold/pgFormatter).

Stop using `sql-formatter` because it doesn't work on the `read_schema` function:
 
```
npx sql-formatter -l postgresql --fix pkg/state/init.sql
````  
doesn't change the file regardless of whatever changes are made to `read_schema`.

`pgFormatter` produces consistent formatting on the `read_schema` function so let's use that.

TBH I'm not convinced that either formatting tool gave us formatting that was better than what we were doing by hand before. It's a question of consistency vs personal preference.


ryanslade:
Agree, it's more about consistency

ryanslade (approved):

Issue #450, opened 2024-11-06
Support for drop default on a column


divyenduz:
The following operation is not supported natively by pgroll. 

```sql
ALTER TABLE public."Pet" ALTER COLUMN id DROP DEFAULT;
```

Current workaround is to use SQL the [escape hatch](https://github.com/xataio/pgroll/blob/main/examples/05_sql.json).

andrew-farries:
Yes, this belongs on `OpAlterColumn`.

I think we could handle this by distinguishing the case where `default` is explicitly set to `null` vs omitted from the operation in the same way as we did in https://github.com/xataio/pgroll/pull/345 for removing comments from tables and columns.

andrew-farries:
The following statements look to have slightly different effects:

```sql
alter table products alter column name set default NULL
```

results in a column that looks like this:

```
 name   | character varying(255) |           |          | NULL::character varying                          | extended |             |              |
```

whereas:

```sql
alter table products alter column name drop default
```

produces a column like this:

```
 name   | character varying(255) |           |          |                                                  | extended |             |              |
```

i.e Postgres catalogs record a default of `NULL` vs no default slightly differently. If we wanted to implement this we'd need to distinguish:

```json
  "operations": [
    {
      "alter_column": {
        "table": "products",
        "column": "name",
        "default": "NULL"
      }
    }
  ]
```

from:

```json
  "operations": [
    {
      "alter_column": {
        "table": "products",
        "column": "name",
        "default": null
      }
    }
  ]
```

WDYT?

ryanslade:
Ok, so we would implement the second case where the value is `null` by dropping the default. I'm still not totally clear on the semantic difference between no default and a default of `NULL`.

andrew-farries:
I don't think there is a difference in practical terms - looks as though Postgres just records each case differently in the system catalogs:

Using this query to retrieve information about column defaults:

```sql
SELECT
     a.attname AS column_name,
     pg_get_expr(ad.adbin, ad.adrelid) AS default_value
 FROM
     pg_catalog.pg_attrdef ad
     JOIN pg_catalog.pg_attribute a ON ad.adnum = a.attnum
 WHERE
     a.attrelid = 'products'::regclass;
```

a `DEFAULT` set to `NULL` still shows up in the result set:

```
+-------------+--------------------------------------------------+
| column_name | default_value                                    |
|-------------+--------------------------------------------------|
| name        | NULL::character varying                          |
+-------------+--------------------------------------------------+
[...]

PR #478, opened 2024-11-21
Support dropping default for a column

ryanslade:
We no allow setting the value of `default` to `null` when altering a column. In practice this will cause us to drop the default. It is also possible to set `default` the *value* `NULL` which has a similar effect but is recorded slightly differently in the PG catalogue. (Details [here](https://github.com/xataio/pgroll/issues/450#issuecomment-2490680272))

Closes https://github.com/xataio/pgroll/issues/450

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -862,7 +862,7 @@ A change default operation changes the default value of a column.
   "alter_column": {
     "table": "table name",
     "column": "column name",
-    "default": "new default value",
+    "default": "new default value (or null)",
     "up": "SQL expression",
     "down": "SQL expression"
   }
@@ -872,6 +872,7 @@ A change default operation changes the default value of a column.
 Example **change default** migrations:
 
 * [35_alter_column_multiple.json](../examples/35_alter_column_multiple.json)
+* [46_alter_column_drop_default](../examples/46_alter_column_drop_default.json)
 
 #### Change comment
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -862,7 +862,7 @@ A change default operation changes the default value of a column.
   "alter_column": {
     "table": "table name",
     "column": "column name",
-    "default": "new default value (or null)",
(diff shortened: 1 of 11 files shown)


pkg/migrations/op_set_default_test.go
```diff
@@ -248,7 +250,109 @@ func TestSetDefault(t *testing.T) {
 						&migrations.OpAlterColumn{
 							Table:   "users",
 							Column:  "name",
-							Default: ptr("NULL"),
+							Default: nullable.NewNullableWithValue("NULL"),
```
ryanslade, line 253: This diff is very confusing. What I actually did was duplicate the test where we set the value to `NULL` with another example where we set it to `null`.

`Default: nullable.NewNullableWithValue("NULL")` vs `Default: nullable.NewNullNullable[string](),`

docs/README.md
```diff
@@ -862,7 +862,7 @@ A change default operation changes the default value of a column.
   "alter_column": {
     "table": "table name",
     "column": "column name",
-    "default": "new default value",
+    "default": "new default value (or null)",
```
andrew-farries, line 865:
```suggestion
    "default": "new default value" | null,
```

examples/46_alter_column_drop_default.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "46_alter_column_drop_default",
+  "operations": [
+    {
+      "alter_column": {
+        "table": "tickets",
+        "column": "ticket_type",
+        "default": null,
+        "up": "",
+        "down": ""
```
andrew-farries, line 10: The data migration in both directions can be a straight copy without modification:

```suggestion
        "up": "ticket_type",
        "down": "ticket_type"
```

andrew-farries (approved):
This looks good, just a couple of small suggestions.

PR #480, opened 2024-11-22
Format examples and add $schema field

ryanslade:


ryanslade:
Closing since the schema itself is not happy with the `$schema` field :)

Issue #473, opened 2024-11-20
Do not require parenthesis `( )` in up/down functions


exekias:
This migration fails cause the up function is not using `()`, let's make it optional so it's less error-prone:

```
[
  {
    "add_column": {
      "table": "posts",
      "column": {
        "name": "userFullName",
        "type": "text",
        "nullable": false
      },
      "up": "select \"fullName\" from users where users.id = userId"
    }
  }
]
```

We should double-check that this change doesn't affect existing migrations or other simpler definitions like a column or static value.

PR #481, opened 2024-11-22
Always surround up and down clauses in parentheses

ryanslade:
Closes https://github.com/xataio/pgroll/issues/473


pkg/migrations/op_add_column_test.go
```diff
@@ -786,6 +786,88 @@ func TestAddColumnWithUpSql(t *testing.T) {
...
+							},
+						},
+					},
+				},
+				{
+					Name: "02_add_column",
+					Operations: migrations.Operations{
+						&migrations.OpAddColumn{
+							Table: "products",
+							Up:    "select name",
```
exekias, line 817: I believe we also want to test just `name` here? ie
```suggestion
							Up:    "name",
```

ryanslade: I'll add a separate test for that. The one with `select` definitely failed without `( )`

ryanslade: Done and looks fine

exekias (approved):

exekias:
probably worth updating docs & examples?

andrew-farries (commented):
I think this is fine. We should update the documentation and example migrations to remove explicit parens as part of this PR too I think.

PR #482, opened 2024-11-22
Handle scan and close errors in SchemaHistory

ryanslade:


pkg/state/history.go
```diff
@@ -33,16 +33,16 @@ func (s *State) SchemaHistory(ctx context.Context, schema string) ([]Migration,
 		return nil, err
 	}
 
-	defer rows.Close()
+	defer func() {
+		err = rows.Close()
+	}()
```
andrew-farries, line 38: Is there a particular reason why we want to return this error? Elsewhere in the codebase we use the simpler

```go
defer someCloser()
```

pattern without assigning to `err` and I think we should be consistent unless there's a reason to deviate.

pkg/state/history.go
```diff
@@ -33,16 +33,16 @@ func (s *State) SchemaHistory(ctx context.Context, schema string) ([]Migration,
...
-	var entries []Migration
 	for rows.Next() {
 		var name, rawMigration string
 		var createdAt time.Time
 
-		rows.Scan(&name, &rawMigration, &createdAt)
-		if err != nil {
-			return nil, err
+		if err := rows.Scan(&name, &rawMigration, &createdAt); err != nil {
+			return nil, fmt.Errorf("row scan: %w", err)
```
andrew-farries, line 45: 👍

pkg/state/history.go
andrew-farries: The simpler style is far more common though. I think unless a cleanup error is especially critical we should follow convention and stick with the one-line pattern.

ryanslade: Actually looks like in this case we can remove `rows.Close`. It is automatically closed when `rows.Next` returns false.

> If [Rows.Next](https://pkg.go.dev/database/sql#Rows.Next) is called and returns false and there are no further result sets, the [Rows](https://pkg.go.dev/database/sql#Rows) are closed automatically and it will suffice to check the result of [Rows.Err](https://pkg.go.dev/database/sql#Rows.Err)

ryanslade: I noticed we don't check `rows.Err` here and probably in many other places too, I'll fix that.

ryanslade:
@andrew-farries PTAL

andrew-farries (approved):

PR #498, opened 2024-11-28
Update docs with deprecation notice

andrew-farries:
Add a deprecation notice to the docs for the `drop_constraint` operation.

Direct link to the updated docs section: [docs/README.md](https://github.com/xataio/pgroll/blob/add-deprecation-note-to-docs/docs/README.md#drop-constraint)

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1210,6 +1210,8 @@ Example **drop column** migrations:
 
 ### Drop constraint
 
+:warning: The **drop constraint** operation is deprecated. Please use the [drop multi-column constraint](#drop-multi-column-constraint) operation instead. The **drop_constraint** operation will be removed in a future release of `pgroll`. :warning:
+
 A drop constraint operation drops a single-column constraint from an existing table.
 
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -1210,6 +1210,8 @@ Example **drop column** migrations:
 
 ### Drop constraint
 
+:warning: The **drop constraint** operation is deprecated. Please use the [drop multi-column constraint](#drop-multi-column-constraint) operation instead. The **drop_constraint** operation will be removed in a future release of `pgroll`. :warning:
```
exekias, line 1213: I wonder if we will rename drop multi column to this once deprecation is in?

exekias (approved):

docs/README.md
andrew-farries: yes, I think so. But I don't think that needs to be part of this notice; it's a breaking change either way.

andrew-farries: unless you think we need to mention that `drop_multicolumn_constraint` will become the new `drop_constraint` here?

PR #499, opened 2024-11-28
Split `docs/README.md` into multiple files

andrew-farries:
Split the content in `docs/README.md` into multiple files under `docs/` for easier reference.


SferaDev (approved):
We can start working with this, let's interate!

PR #502, opened 2024-12-03
Add a `sql2pgroll` package to convert SQL to `pgroll` migrations

andrew-farries:
Add a `sql2pgroll` package to convert SQL to `pgroll` migrations.

Add a (hidden for now) `pgroll sql` command that uses the package to convert SQL strings on the command line to `pgroll` migrations.

The `sql2pgroll` package is incomplete, with almost all SQL falling back to conversion using raw SQL migrations. Only some `CREATE TABLE` statements and the `ALTER TABLE ... ALTER COLUMN ... SET NOT NULL` statement are currently handled.

```bash
$ pgroll sql "create table foo(a serial primary key, b text unique)" 
```

```json
[
  {
    "create_table": {
      "columns": [
        {
          "name": "a",
          "pk": true,
          "type": "serial"
        },
        {
          "name": "b",
          "nullable": true,
          "type": "text",
          "unique": true
        }
      ],
      "name": "foo"
    }
  }
]
```

Part of #504

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -79,6 +79,7 @@ func Execute() error {
 	rootCmd.AddCommand(migrateCmd())
 	rootCmd.AddCommand(pullCmd())
 	rootCmd.AddCommand(latestCmd())
+	rootCmd.AddCommand(sqlCmd())
 
 	return rootCmd.Execute()
 }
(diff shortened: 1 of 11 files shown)


andrew-farries:
Issue tracking the work remaining on `sql2pgroll`: https://github.com/xataio/pgroll/issues/504

exekias (approved):
awesome feature! looking forward to seing it progressing.

On the command name, I'm unsure about `sql` but don't have a better proposal, so let's go for it and see how it goes

PR #518, opened 2024-12-09
Use `ubuntu 24.04` as the runner in all workflows

andrew-farries:
Update the runner used by all jobs in all workflows to use `ubuntu-24.04` instead of `ubuntu-latest`.

`ubuntu-24.04` will become `ubuntu-latest` over the next month, see:

https://github.com/actions/runner-images/issues/10636

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   benchmark:
     name: 'benchmark (pg: ${{ matrix.pgVersion }})'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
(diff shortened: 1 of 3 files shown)


andrew-farries:
@ryanslade this may impact the benchmark workflow. Would you rather the `benchmark` workflow continued to use `ubuntu-22.04` ?

SferaDev (approved):

ryanslade:
> @ryanslade this may impact the benchmark workflow. Would you rather the `benchmark` workflow continued to use `ubuntu-22.04` ?

Are you worried it may change the results? I think that's ok since we will know the cause. Also, I think it's better to fix the version so we control when it changes and can correlate any performance changes.

Issue #504, opened 2024-12-03
Implement `sql2pgroll` package


andrew-farries:
Support for converting SQL DDL statements to `pgroll` migrations is partially complete. This issue tracks progress towards completion.

## `pgroll` operations

| Operation                        | Status      | Notes                   |
|----------------------------------|-------------|-------------------------|
| `create_table`                   |      🟠        |                         |
| `rename_table`                   |    🔴          |                         |
| `drop_table`                     |      🔴        |                         |
| `add_column`                     |     🔴         |                         |
| `drop_column`                    |     🔴         |                         |
| `alter_column` (rename column)                  |  🔴            |                         |
| `alter_column` (change type)                    |      🔴        |                         |
| `alter_column` (set check constraint)           |     🔴         |                         |
| `alter_column` (set not null)                   |      🟢         |                         |
| `alter_column` (unset not null)                   |      🔴          |                         |
| `alter_column` (set foreign key)                |       🔴       |                         |
| `alter_column` (set_unique)                     |     🔴         |                         |
| `alter_column` (set comment) |     🔴               | 
| `alter_column` (set default) |     🔴               | 
| `alter_column` (drop default) |     🔴               | 
| `create_index`                   |       🔴       |                         |
| `drop_index`                     |       🔴       |                         |
| `rename_constraint`              |      🔴        |                         |
| `drop_constraint`                |        🔴      |   op is deprecated, `sql2pgroll` should write only `drop_multicolumn_constraint` ops                      |
| `set_replica_identity`           |      🔴        |                         |
| `drop_multicolumn_constraint`    |    🔴          |                         |
| `create_constraint`              |     🔴         |                         |

andrew-farries:
I'm going to close this as complete now that we have implemented `sql2pgroll` for existing operation types. 

cc @kvch

PR #521, opened 2024-12-10
Convert DROP COLUMN SQL to pgroll operation

ryanslade:


ryanslade:
@andrew-farries What should we do about more complicated statements? The fun spec is:

```sql
DROP [ COLUMN ] [ IF EXISTS ] column_name [ RESTRICT | CASCADE ]
```

pkg/sql2pgroll/convert.go
```diff
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	pgq "github.com/pganalyze/pg_query_go/v6"
+
```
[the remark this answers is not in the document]
ryanslade: I'm using Goland which tries to sort the imports into three groups:

standard library
third party
Imports starting with `github.com/xataio/pgroll`

ryanslade:
@andrew-farries Thanks for the help, this is ready for review.

pkg/sql2pgroll/alter_table.go
```diff
@@ -155,6 +158,32 @@ func convertAlterTableAddUniqueConstraint(stmt *pgq.AlterTableStmt, constraint *
...
+		// Supported
+	case pgq.DropBehavior_DROP_CASCADE:
+		// Fall back to SQL
+		return nil, nil
+	}
+
+	// IF EXISTS not supported
+	if cmd.MissingOk {
+		return nil, nil
+	}
```
andrew-farries, line 178: For other operations we have the pattern of pulling out these checks into `canConvert` functions:

https://github.com/xataio/pgroll/blob/main/pkg/sql2pgroll/alter_table.go?plain=1#L161-L188

It's up to you if you want to follow the pattern here.

pkg/sql2pgroll/alter_table_test.go
```diff
@@ -72,6 +81,10 @@ func TestUnconvertableAlterTableAddConstraintStatements(t *testing.T) {
 		// operations when changing data type.
 		`ALTER TABLE foo ALTER COLUMN a SET DATA TYPE text COLLATE "en_US"`,
 		"ALTER TABLE foo ALTER COLUMN a SET DATA TYPE text USING 'foo'",
+
+		// CASCADE and IF EXISTS clauses are not represented by OpDropColumn
+		"ALTER TABLE foo DROP COLUMN bar CASCADE",
+		"ALTER TABLE foo DROP COLUMN IF EXISTS bar",
```
andrew-farries, line 87: We should rename the test here now that it's not just covering unconvertible add constraint operations.

ryanslade: Done

pkg/sql2pgroll/alter_table.go
ryanslade: Done

ryanslade:
Once merged, do we mark the entry in the table at https://github.com/xataio/pgroll/issues/504 with orange or green? We still fall back to raw sql in some cases so not sure if that is considered "done done"

andrew-farries:
> Once merged, do we mark the entry in the table at #504 with orange or green? We still fall back to raw sql in some cases so not sure if that is considered "done done"

I think we can set it to 🟢 . The 'fallback to raw sql' testcases serve as a good list of extra options for the respective `pgroll` operations to support, but we don't need to commit to supporting all of them as part of #504.

andrew-farries (approved):

PR #524, opened 2024-12-11
Convert DROP INDEX SQL to pgroll operation

ryanslade:
Converts SQL in the following forms to the equivalent pgroll operation:

```sql
DROP INDEX foo
DROP INDEX schema.foo
DROP INDEX foo RESTRICT
DROP INDEX CONCURRENTLY foo
DROP INDEX IF EXISTS foo
```

The following forms are left as raw SQL operations since we do not support them in pgroll yet:

```sql
DROP INDEX foo CASCADE
```

### DIFF
--- a/pkg/sql2pgroll/drop.go
+++ b/pkg/sql2pgroll/drop.go
@@ -3,6 +3,8 @@
 package sql2pgroll
 
 import (
+	"strings"
+
 	pgq "github.com/pganalyze/pg_query_go/v6"
 
 	"github.com/xataio/pgroll/pkg/migrations"
@@ -21,10 +23,15 @@ func convertDropIndexStatement(stmt *pgq.DropStmt) (migrations.Operations, error
 	if !canConvertDropIndex(stmt) {
 		return nil, nil
 	}
-	s := stmt.GetObjects()[0].GetList().GetItems()[0].GetString_().GetSval()
+	items := stmt.GetObjects()[0].GetList().GetItems()
+	parts := make([]string, len(items))
+	for i, item := range items {
+		parts[i] = item.GetString_().GetSval()
+	}
+
 	return migrations.Operations{
 		&migrations.OpDropIndex{
-			Name: s,
+			Name: strings.Join(parts, "."),
 		},
 	}, nil
 }
--- a/pkg/sql2pgroll/drop.go
+++ b/pkg/sql2pgroll/drop.go
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package sql2pgroll
+
+import (
+	pgq "github.com/pganalyze/pg_query_go/v6"
+
+	"github.com/xataio/pgroll/pkg/migrations"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/sql2pgroll/drop.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	return nil, nil
+}
+
+// convertDropIndexStatement converts simple DROP INDEX statements to pgroll operations
+func convertDropIndexStatement(stmt *pgq.DropStmt) (migrations.Operations, error) {
+	if !canConvertDropIndex(stmt) {
+		return nil, nil
+	}
+	s := stmt.Objects[0].GetList().Items[0].GetString_().GetSval()
```
andrew-farries, line 24: I think this is going to fail in the case of schema-qualified index drops:

```sql
DROP INDEX myschema.idx_foo
```

pkg/sql2pgroll/drop.go
```diff
@@ -0,0 +1,45 @@
...
+		},
+	}, nil
+}
+
+// canConvertDropIndex checks whether we can convert the statement without losing any information.
+func canConvertDropIndex(stmt *pgq.DropStmt) bool {
+	if len(stmt.Objects) > 1 {
+		return false
+	}
+	if stmt.MissingOk || stmt.Concurrent {
```
andrew-farries, line 37: `pgroll` always drops indexes concurrently (see `op_drop_index.go`).

Conversion should succeed whether the `CONCURRENTLY` option is set or not.

pkg/sql2pgroll/drop_test.go
```diff
@@ -0,0 +1,64 @@
...
+			assert.Equal(t, tc.expectedOp, ops[0])
+		})
+	}
+}
+
+func TestUnconvertableDropIndexStatements(t *testing.T) {
+	t.Parallel()
+
+	tests := []string{
+		"DROP INDEX CONCURRENTLY foo",
```
andrew-farries, line 49: This testcase should be removed, see the other comment about always dropping indexes concurrently.

pkg/sql2pgroll/drop.go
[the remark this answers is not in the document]
andrew-farries: We can use accessor methods consistently (`GetObjects`, `GetItems`) but I don't think there is a much better way to do it.

andrew-farries (commented):
Looks good 👍 , left some comments

pkg/sql2pgroll/drop.go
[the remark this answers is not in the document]
ryanslade: Ok, done

pkg/sql2pgroll/drop.go
ryanslade: I see we also use `IF EXISTS`

andrew-farries: yes, I think we should convert regardless of whether `IF EXISTS` is specified or not.

ryanslade:
@andrew-farries PTAL

PR #526, opened 2024-12-11
sql2pgroll: Support set and drop column defaults

ryanslade:
Support translating the following alter table statements into pgroll operations:

```sql
ALTER TABLE foo COLUMN bar SET DEFAULT 'foo'
ALTER TABLE foo COLUMN bar SET DEFAULT 123
ALTER TABLE foo COLUMN bar SET DEFAULT 123.456
ALTER TABLE foo COLUMN bar SET DEFAULT true
ALTER TABLE foo COLUMN bar SET DEFAULT B'0101'
ALTER TABLE foo COLUMN bar SET DEFAULT null
ALTER TABLE foo COLUMN bar DROP DEFAULT
```

For cases where the default is set to something other than a simple literal, we fall back to raw SQL.

### DIFF
--- a/pkg/sql2pgroll/alter_table.go
+++ b/pkg/sql2pgroll/alter_table.go
@@ -5,6 +5,7 @@ package sql2pgroll
 import (
 	"fmt"
 
+	"github.com/oapi-codegen/nullable"
 	pgq "github.com/pganalyze/pg_query_go/v6"
 
 	"github.com/xataio/pgroll/pkg/migrations"
@@ -38,6 +39,8 @@ func convertAlterTableStmt(stmt *pgq.AlterTableStmt) (migrations.Operations, err
 			op, err = convertAlterTableAddConstraint(stmt, alterTableCmd)
 		case pgq.AlterTableType_AT_DropColumn:
 			op, err = convertAlterTableDropColumn(stmt, alterTableCmd)
+		case pgq.AlterTableType_AT_ColumnDefault:
+			op, err = convertAlterTableSetColumnDefault(stmt, alterTableCmd)
 		}
 
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/sql2pgroll/alter_table.go
```diff
@@ -158,11 +161,27 @@ func convertAlterTableAddUniqueConstraint(stmt *pgq.AlterTableStmt, constraint *
...
+// convertAlterTableSetColumnDefault converts SQL statements like:
 //
-// `ALTER TABLE foo DROP COLUMN bar
+// `ALTER TABLE foo COLUMN bar SET DEFAULT 'foo'`
+// `ALTER TABLE foo COLUMN bar SET DEFAULT null`
+// `ALTER TABLE foo COLUMN bar DROP DEFAULT`
 //
-// to an OpDropColumn operation.
+// to an OpAlterColumn operation.
+func convertAlterTableSetColumnDefault(stmt *pgq.AlterTableStmt, cmd *pgq.AlterTableCmd) (migrations.Operation, error) {
```
ryanslade, line 171: Since `OpAlterColumn` only supports nullable string defaults, this will fail for any other kind of default for example non strings or the value of a function call.

In those cases should we fall back to raw sql?

ryanslade: It looks like if the new default value is a constant, we can support these types:

```go
type A_Const struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	// Types that are assignable to Val:
	//
	//	*A_Const_Ival
	//	*A_Const_Fval
	//	*A_Const_Boolval
	//	*A_Const_Sval
	//	*A_Const_Bsval
	Val      isA_Const_Val `protobuf_oneof:"val"`
	Isnull   bool          `protobuf:"varint,10,opt,name=isnull,proto3" json:"isnull,omitempty"`
	Location int32         `protobuf:"varint,11,opt,name=location,proto3" json:"location,omitempty"`
}
```

And then convert them to a string

ryanslade: I've done this

andrew-farries (commented):
This is a good start 👍 

This won't work for all types of column default though:

```sql
-- integers
ALTER TABLE foo ALTER COLUMN bar SET DEFAULT 2
-- booleans
ALTER TABLE foo ALTER COLUMN bar SET DEFAULT true
-- floating point values
ALTER TABLE foo ALTER COLUMN bar SET DEFAULT 2.0
-- bitstrings
ALTER TABLE foo ALTER COLUMN bar SET DEFAULT B'10101010'
-- function calls
ALTER TABLE foo ALTER COLUMN bar SET DEFAULT NOW()
```

These cases all have a different AST structure and need to be handled separately. In general a column `DEFAULT` can be an expression of arbitrary complexity, such as:

```sql
ALTER TABLE foo ALTER COLUMN bar SET DEFAULT (CASE WHEN EXTRACT(DOW FROM CURRENT_TIMESTAMP) IN (0, 6) THEN 1 ELSE 5 END)
```

In the latter case I think we need to make changes to the `pg_query_go` dependency to support deparsing arbitrary SQL expressions, rather than only the full SQL statements that it currently supports.

For now we should aim to convert `DEFAULT` expressions where the expression is a simple literal:
* integer
* boolean
* floating point
* bit string

We can leave handling function calls and more complex SQL until later.

ryanslade:
Cool, that's exactly what I did 😄 

I'll throw in a test case where it's a function value, for example `now()` and fall back to raw SQL in that case.

ryanslade:
@andrew-farries Ok, I think I've handled all the cases now.

andrew-farries (approved):

PR #527, opened 2024-12-11
Build `CGO_ENABLED` binaries for supported platforms in the `release` job

andrew-farries:
Make the `release` job run in a `goreleaser/goreleaser-cross` container so that we can build binaries for all supported architectures.

See the PR description for https://github.com/xataio/pgroll/pull/525 for details of why we need a cross-compilation toolchain as a result of taking a dependency on `pg_query_go` and how the the `goreleaser-cross` image is used.

The addition of the `-static` `LDFLAG` for the `linux/amd64` and `linux/arm64` builds is necessary to statically link in order for the resulting `pgroll` binaries to run in the `FROM scratch` image that `Goreleaser` uses to build docker images.

Also, revert #515 to re-add the `pgroll sql`  command now that `pgroll` can be built for all platforms when the dependency on `pg_query_go` is included in the build.

### DIFF
--- a/cmd/sql.go
+++ b/cmd/sql.go
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	"github.com/xataio/pgroll/pkg/sql2pgroll"
+)
+
+func sqlCmd() *cobra.Command {
+	sqlCmd := &cobra.Command{
+		Use:    "sql <sql statement>",
+		Short:  "Convert SQL statements to pgroll operations",
+		Args:   cobra.ExactArgs(1),
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


cmd/sql.go
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
```
ryanslade, line 1: Did you mean to commit this? It seems unrelated to CI

andrew-farries: Yes, this is re-adding the `pgroll sql` command that we had to remove in https://github.com/xataio/pgroll/pull/515.

We had to remove it because it caused [pgquery_go](https://github.com/pganalyze/pg_query_go) to be compiled as part of the project. `pgquery_go` uses `cgo` and so made compilation for different platforms more complicated than when `pgroll` was a pure-Go binary.

By removing this file the only dependency on `pgquery_go` was in tests. Now that we can compile with `CGO_ENABLED` on each of the platforms we need to target we can re-introduce the dependency.

ryanslade (approved):

PR #536, opened 2024-12-16
Convert DROP CONSTRAINT SQL into OpDropMultiColumnConstraint operations

ryanslade:
Convert `DROP CONSTRAINT SQL` into an `OpDropMultiColumnConstraint`.
Because we are unable to infer the columns involved, placeholder migrations are used.

SQL statements like the following are supported:

```sql
ALTER TABLE foo DROP CONSTRAINT constraint_foo
ALTER TABLE foo DROP CONSTRAINT IF EXISTS constraint_foo
ALTER TABLE foo DROP CONSTRAINT IF EXISTS constraint_foo RESTRICT
```

`CASCADE` is currently not supported and will fall back to raw SQL

### DIFF
--- a/pkg/sql2pgroll/alter_table.go
+++ b/pkg/sql2pgroll/alter_table.go
@@ -42,6 +42,8 @@ func convertAlterTableStmt(stmt *pgq.AlterTableStmt) (migrations.Operations, err
 			op, err = convertAlterTableDropColumn(stmt, alterTableCmd)
 		case pgq.AlterTableType_AT_ColumnDefault:
 			op, err = convertAlterTableSetColumnDefault(stmt, alterTableCmd)
+		case pgq.AlterTableType_AT_DropConstraint:
+			op, err = convertAlterTableDropConstraint(stmt, alterTableCmd)
 		}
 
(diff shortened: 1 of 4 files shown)


pkg/sql2pgroll/alter_table.go
```diff
@@ -293,6 +295,37 @@ func convertAlterTableSetColumnDefault(stmt *pgq.AlterTableStmt, cmd *pgq.AlterT
...
+
+	return &migrations.OpDropMultiColumnConstraint{
+		Up: migrations.MultiColumnUpSQL{
+			"placeholder": PlaceHolderSQL,
+		},
+		Down: migrations.MultiColumnDownSQL{
+			"placeholder": PlaceHolderSQL,
+		},
+		Table: stmt.GetRelation().GetRelname(),
+		Name:  cmd.GetName(),
```
andrew-farries, line 321: I think we miss schema-qualified names here.

ryanslade: Ah, good catch

andrew-farries (approved):

PR #540, opened 2024-12-17
Use a helper function to get a fully qualified relation name

ryanslade:


pkg/sql2pgroll/alter_table.go
```diff
@@ -443,6 +430,17 @@ func canConvertColumnForSetDataType(column *pgq.ColumnDef) bool {
 	return true
 }
 
+func getQualifiedRelationName(rel interface {
+	GetSchemaname() string
+	GetRelname() string
+},
```
andrew-farries, line 436: Do we need the generality of the interface here? All the call sites pass in a `*pgq.RangeVar` - maybe the parameter could be of type `*pgq.RangeVar` instead?

andrew-farries (commented):
Left a comment.

Also, maybe the `gofumpt` stuff could go in a separate PR?

pkg/sql2pgroll/alter_table.go
ryanslade: Ah, yeah, good point. I'll make those changes.

andrew-farries (approved):

PR #552, opened 2024-12-19
Convert SQL `CREATE TABLE` inline `CHECK` constraints

andrew-farries:
Convert SQL `CREATE TABLE` statements with inline `CHECK` constraints like this:

```
CREATE TABLE foo(a int CHECK (a > 0))
```

to `OpCreateTable` operations like this:

```json
[
  {
    "create_table": {
      "columns": [
        {
          "check": {
            "constraint": "a > 0",
            "name": "foo_a_check"
          },
          "name": "a",
          "nullable": true,
          "type": "int"
        }
      ],
      "name": "foo"
    }
  }
]
```

### DIFF
--- a/pkg/sql2pgroll/create_table.go
+++ b/pkg/sql2pgroll/create_table.go
@@ -97,6 +97,7 @@ func convertColumnDef(col *pgq.ColumnDef) (*migrations.Column, error) {
 
 	// Convert column constraints
 	var notNull, pk, unique bool
+	var check *migrations.CheckConstraint
 	for _, c := range col.GetConstraints() {
 		switch c.GetConstraint().GetContype() {
 		case pgq.ConstrType_CONSTR_NULL:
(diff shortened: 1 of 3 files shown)


pkg/sql2pgroll/create_table.go
```diff
@@ -41,7 +41,7 @@ func convertCreateStmt(stmt *pgq.CreateStmt) (migrations.Operations, error) {
 
 	return migrations.Operations{
 		&migrations.OpCreateTable{
-			Name:    stmt.Relation.Relname,
+			Name:    stmt.Relation.GetRelname(),
```
ryanslade, line 44: Do we need to support qualified names here?

ryanslade (approved):

pkg/sql2pgroll/create_table.go
andrew-farries: Thanks 🙏 I fixed it with a separate PR: https://github.com/xataio/pgroll/pull/557

PR #553, opened 2024-12-19
Convert column `DEFAULT`s in `CREATE TABLE` statements

andrew-farries:
Convert inline `DEFAULT` values in `CREATE TABLE` statements like this:

```sql
CREATE TABLE foo(a text DEFAULT 'foo')
```

into `OpCreateTable` operations like this:

```json
[
  {
    "create_table": {
      "columns": [
        {
          "default": "'foo'",
          "name": "a",
          "nullable": true,
          "type": "text"
        }
      ],
      "name": "foo"
    }
  }
]
```

### DIFF
--- a/pkg/sql2pgroll/create_table_test.go
+++ b/pkg/sql2pgroll/create_table_test.go
@@ -45,6 +45,10 @@ func TestConvertCreateTableStatements(t *testing.T) {
 			expectedOp: expect.CreateTableOp10,
 		},
 		{
+			sql:        "CREATE TABLE foo(a timestamptz DEFAULT now())",
+			expectedOp: expect.CreateTableOp11,
+		},
+		{
 			sql:        "CREATE TABLE foo(a varchar(255))",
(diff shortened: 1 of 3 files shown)


pkg/sql2pgroll/create_table_test.go
```diff
@@ -44,6 +44,10 @@ func TestConvertCreateTableStatements(t *testing.T) {
 			sql:        "CREATE TABLE foo(a int CHECK (a > 0))",
 			expectedOp: expect.CreateTableOp10,
 		},
+		{
```
ryanslade, line 47: Should we add a test for a non-constant default like `now()`?

ryanslade (approved):

pkg/sql2pgroll/create_table_test.go
andrew-farries: I just changed the testcase to use a `now()` as you suggest; I don't think we need two testcases here.

PR #554, opened 2024-12-19
Convert column `FOREIGN KEY`s in `CREATE TABLE`

andrew-farries:
Convert inline `FOREIGN KEY` constraints in `CREATE TABLE` statements like this:

```sql
CREATE TABLE foo(a int REFERENCES bar(b))
```

into `OpCreateTable` operations like this:

```json
[
  {
    "create_table": {
      "columns": [
        {
          "name": "a",
          "nullable": true,
          "references": {
            "column": "b",
            "name": "foo_a_fkey",
            "on_delete": "NO ACTION",
            "table": "bar"
          },
          "type": "int"
        }
      ],
      "name": "foo"
    }
  }
]
```

### DIFF
--- a/pkg/sql2pgroll/expect/create_table.go
+++ b/pkg/sql2pgroll/expect/create_table.go
@@ -129,3 +129,88 @@ var CreateTableOp11 = &migrations.OpCreateTable{
 		},
 	},
 }
+
+var CreateTableOp12 = &migrations.OpCreateTable{
+	Name: "foo",
+	Columns: []migrations.Column{
+		{
+			Name:     "a",
+			Type:     "int",
+			Nullable: true,
+			References: &migrations.ForeignKeyReference{
+				Name:     "foo_a_fkey",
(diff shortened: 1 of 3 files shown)


pkg/sql2pgroll/expect/create_table.go
```diff
@@ -129,3 +129,88 @@ var CreateTableOp11 = &migrations.OpCreateTable{
 		},
 	},
 }
+
+var CreateTableOp12 = &migrations.OpCreateTable{
```
ryanslade, line 133: Since all of these only differ based on the `OnDelete` action maybe a helper method would be useful?

ryanslade (approved):

pkg/sql2pgroll/expect/create_table.go
andrew-farries: yes, will do this separately 👍

PR #559, opened 2024-12-19
Allow some named inline column constraints in `CREATE TABLE` statements

andrew-farries:
Allow some named inline column constraints in `CREATE TABLE` statements.

SQL like this can be converted to a pgroll `OpCreateTable`:

```sql
CREATE TABLE foo(a int CONSTRAINT my_check CHECK (a > 0))
CREATE TABLE foo(a int CONSTRAINT my_fk REFERENCES bar(b))
```

The previous behaviour was to allow only the unnamed forms of these constraints:

```sql
CREATE TABLE foo(a int CHECK (a > 0))
CREATE TABLE foo(a int REFERENCES bar(b))
```

Named inline column constraints remain unsupported for `DEFAULT`, `NULL`, `NOT NULL`, `UNIQUE` and `PRIMARY KEY` constraints and will fall back to a raw SQL operation.

### DIFF
--- a/pkg/sql2pgroll/create_table.go
+++ b/pkg/sql2pgroll/create_table.go
@@ -4,7 +4,6 @@ package sql2pgroll
 
 import (
 	"fmt"
-	"slices"
 
 	pgq "github.com/xataio/pg_query_go/v6"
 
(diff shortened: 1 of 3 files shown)


pkg/sql2pgroll/create_table.go
```diff
@@ -254,3 +265,15 @@ func convertInlineForeignKeyConstraint(tableName, columnName string, constraint
 		Table:    getQualifiedRelationName(constraint.GetPktable()),
 	}, nil
 }
+
+// isConstraintNamed returns true iff `constraint` has a name.
+// Column constraints defined inline in CREATE TABLE statements cab be either
```
ryanslade, line 270:
```suggestion
// Column constraints defined inline in CREATE TABLE statements can be either
```

ryanslade (approved):

PR #567, opened 2025-01-03
Show completion estimate during backfill

### LINKED ISSUE #492: Include percentage complete information during migration start
On migration start, `pgroll` currently provides progress updates in terms of the number of rows backfilled:

https://github.com/xataio/pgroll/assets/8225907/75b67093-719b-4c92-bcf2-2faacca7eef7

`pgroll` should also include at least an estimate of the total number of rows in the table being backfilled.

One approach to this would be to use row estimates from system catalogs, to avoid running `COUNT(*)` on large tables, falling back to a `COUNT(*)` only if the estimate is 0.

ryanslade:
Instead of only showing the number of rows backfills, show an estimate of the total number of tows completed as a percentage.

For example:

```
1500 records complete... (12.34%)
```

It attempts to estimate the total number of rows but will fall back to a full scan if the number of rows estimated is zero.

Closes #492

### DIFF
--- a/pkg/db/db.go
+++ b/pkg/db/db.go
@@ -20,6 +20,7 @@ const (
 
 type DB interface {
 	ExecContext(ctx context.Context, query string, args ...interface{}) (sql.Result, error)
+	QueryRowContext(ctx context.Context, query string, args ...interface{}) *sql.Row
 	WithRetryableTransaction(ctx context.Context, f func(context.Context, *sql.Tx) error) error
 	Close() error
 }
@@ -50,6 +51,11 @@ func (db *RDB) ExecContext(ctx context.Context, query string, args ...interface{
 
 		return nil, err
 	}
+}
+
+// QueryRowContext wraps sql.DB.QueryRowContext.
+func (db *RDB) QueryRowContext(ctx context.Context, query string, args ...interface{}) *sql.Row {
+	return db.DB.QueryRowContext(ctx, query, args...)
 }
 
 // WithRetryableTransaction runs `f` in a transaction, retrying on lock_timeout errors.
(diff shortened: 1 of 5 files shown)


pkg/db/db.go
```diff
@@ -52,6 +53,11 @@ func (db *RDB) ExecContext(ctx context.Context, query string, args ...interface{
 	}
 }
 
+// QueryRowContext wraps sql.DB.QueryRowContext.
+func (db *RDB) QueryRowContext(ctx context.Context, query string, args ...interface{}) *sql.Row {
+	return db.DB.QueryRowContext(ctx, query, args...)
```
ryanslade, line 58: Errors are deferred until scan so we can't check for transaction timeouts here

andrew-farries: I think either;

* Use `QueryContext` and assume one row and make it retry on lock timeouts.
* Create a wrapper interface for `*sql.Row` like we did for `*sql.DB` and then have a `Scan` method that retries on lock timeouts.* Add a `GetDB()` method to the `DB` interface to return the underlying `*sql.DB` and call `QueryRowContext` on that - that avoids setting the false expectation that the query will be retried on lock timeouts.

andrew-farries (commented):
Looks good, just left a comment about the retryable query.

pkg/db/db.go
ryanslade: I'll go with the first option as it the most consistent with what we currently have.

ryanslade:
@andrew-farries PTAL

andrew-farries (commented):
This looks good 👍 

We have tests for this package in `db_test.go`; we should add tests for `QueryContext` just as we have for `ExecContext`:

https://github.com/xataio/pgroll/blob/d64ba26a54eea15d805dcd31d1a3b1582fbccabc/pkg/db/db_test.go?plain=1#L22-L62

Issue #571, opened 2025-01-07
pgroll prefixes remain on index names after migration is completed


agedemenli:
For CREATE TABLE operation,
Example below:
```
postgres=# \d+ customers
                                                Table "public.customers"
 Column │          Type          │ Collation │ Nullable │ Default │ Storage  │ Compression │ Stats target │ Description
────────┼────────────────────────┼───────────┼──────────┼─────────┼──────────┼─────────────┼──────────────┼─────────────
 id     │ integer                │           │ not null │         │ plain    │             │              │
 name   │ character varying(255) │           │ not null │         │ extended │             │              │
 bio    │ text                   │           │          │         │ extended │             │              │
Indexes:
    "_pgroll_new_customers_pkey" PRIMARY KEY, btree (id)
    "_pgroll_new_customers_name_key" UNIQUE CONSTRAINT, btree (name)
Access method: heap
```

I guess it's not because we name them like this, but instead because they automatically get their names from the table name. We should either consider renaming them after all is done; or maybe just go with the original name, as there's no point for using a temporary name when creating a table from scratch, so that indexes will get their names without the prefix.

PR #572, opened 2025-01-07
Use original table name without prefix when processing create table

agedemenli:
Using the original name, instead of calling `TemporaryName` which adds the prefix `_pgroll_new_` in create table operations and also when rolling back.
Includes necessary changes in tests.
Fixes: #571


pkg/migrations/op_create_table.go
```diff
@@ -22,10 +22,9 @@ func (o *OpCreateTable) Start(ctx context.Context, conn db.DB, latestSchema stri
 		return nil, fmt.Errorf("failed to create columns SQL: %w", err)
 	}
 
-	// Create the table under a temporary name
-	tempName := TemporaryName(o.Name)
+	// Create the table using the original name
```
andrew-farries, line 25:
```suggestion
	// Create the table
```

andrew-farries (approved):
This looks great 👍

PR #574, opened 2025-01-08
Use pointers for `schema`  and `Table` struct fields

andrew-farries:
Update the `schema` and `Tables` structs in the `schema` package to use pointer types in fields.

As part of #239, the in-memory schema representation will need to be updated more often and more accurately by individual operations in a migration. Using pointers in these `schema` package structs makes this possible.


kvch (approved):

exekias (approved):

kvch:
Just to be clear, it means that we will not maintain a separate virtual schema for in progress operations? Instead everything will be stored in the same schema, not just the physical schema?

andrew-farries:
> Just to be clear, it means that we will not maintain a separate virtual schema for in progress operations? Instead everything will be stored in the same schema, not just the physical schema?

Yes, that's right.

The virtual schema is updated by each operation in a migration to reflect the changes made by that operation. I don't think we need a separate virtual schema for in-progress operations.

In practice, the only changes that need to be made to the virtual schema by each operation are those changes that involve creating indirections. For example, when an `OpRenameTable` renames a table from `foo` to `bar` the virtual schema is updated to look like:

```json
{
  "tables": {
    "bar": {
      "name": "foo",
      ...
    },
    ...
  }
}
```

Likewise if a column is renamed by an `OpAlterColumn` from `foo` to `bar` the virtual schema is updated to look like:

```json
{
  "tables": {
    "some_table": {
      "name": "some_table",
      "columns": {
        "bar": { "name": "foo",  ...},
      },
      ...
    },
    ...
  }
}
```

This is because ultimately, the only use for the virtual schema is to be able to create the views in the version schema at the end of the migration for which precise details about changes to column types and constraints etc is not needed.

PR #586, opened 2025-01-13
Rollback migration operations in reverse order

andrew-farries:
Ensure that operations within a migration are rolled back in reverse order when rolling a back a multi-operation migration.

To make this possible, rollbacks now work like this:

On migration rollback:
* Obtain the previous version of the schema (eg, if we are rolling back `02_bar`, read the schema that resulted from the application of `01_foo`). This uses the new `state.SchemaAfterMigration` method.
* Replay the migration `02_bar` on this schema, using a fake DB connection to build up the changes to the virtual schema `s` without altering the physical schema. This ensures that any indirections and temporary columns are present in the virtual schema `s`.
* Invoke `Rollback` on each operation in reverse order, passing `s` as the schema argument. Each `Rollback` may then update `s` to undo the changes to the virtual schema made by the operation, ensuring that preceding operations in the migration see the virtual schema as it was when the operation was applied.


kvch (approved):

kvch (commented):
This is an improvement compared to the previous behaviour, so I approved it. However, I am wondering what happens if e.g. we have 3 operations in a migration and the first one succeeds, the last two fails and are never executed. We still try to rollback all 3 operations, right? Let's say that operation 2 is creating a table, and operation 3 is adding a constraint. On rollback `pgroll` tries to remove the constraint using `ALTER TABLE` on the non existent table, so the rollback will fail.

We should track what operations have to be rolled back in a migration file. Also, we must make `Rollback` functions more robust by checking the schema before running SQL statements. I assume we can mitigate the issue I raised by checking if the table exists in the virtual schema in the `Rollback` function of `OpCreateConstraint`.

andrew-farries:
> We still try to rollback all 3 operations, right? Let's say that operation 2 is creating a table, and operation 3 is adding a constraint. On rollback pgroll tries to remove the constraint using ALTER TABLE on the non existent table, so the rollback will fail.

Yes, I think that's correct as things stand today- all operations in the migration will try to roll back. We can mitigate problems here by making rollbacks no-op in case the object they need to act on doesn't exist.

> we must make `Rollback` functions more robust by checking the schema before running SQL statements. I assume we can mitigate the issue I raised by checking if the table exists in the virtual schema in the `Rollback` function of `OpCreateConstraint`.

Yes, that's what the rest of #239 is about - ensuring that each operation type updates the virtual schema during `Start`, `Rollback` and `Validate` so that operations in the same migration are aware of changes made by the other operations.

PR #587, opened 2025-01-13
Multi-operation migration support for `drop_table` operations

andrew-farries:
Ensure that `drop_table` operations can be used as part of multi-operation migrations:

Add testcases for:
* **Create table, rename table, drop table**: 
  * Make `OpRenameTable.Validate` update the in-memory schema so that the following operation can validate using the new name of the table
  * Make`OpDropTable` soft delete the table from the virtual schema, so that `OpDropTable.Rollback` can un-drop the table, making it visible to preceding operations in the table when the migration is rolled back.
  * Make `OpRenameTable` undo the table rename, so that the table is visible under its old name for preceding operations in the rollback.
* **Drop table, rename table** and **drop table, drop table**:
  * Make `OpDropTable.Validate` remove the table from the virtual schema so that validation of the subsequent operation fails.

Theses changes ensure that `drop_table` operations can be used as part of multi-operation migrations.

### DIFF
--- a/pkg/migrations/op_drop_table.go
+++ b/pkg/migrations/op_drop_table.go
@@ -37,5 +37,7 @@ func (o *OpDropTable) Validate(ctx context.Context, s *schema.Schema) error {
 	if table == nil {
 		return TableDoesNotExistError{Name: o.Name}
 	}
+
+	s.RemoveTable(table.Name)
 	return nil
 }
--- a/pkg/migrations/op_drop_table.go
+++ b/pkg/migrations/op_drop_table.go
@@ -25,6 +25,9 @@ func (o *OpDropTable) Complete(ctx context.Context, conn db.DB, tr SQLTransforme
 }
 
 func (o *OpDropTable) Rollback(ctx context.Context, conn db.DB, tr SQLTransformer, s *schema.Schema) error {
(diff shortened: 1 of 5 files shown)


exekias:
what will happen if you do a drop_table + create_table with the same name? I guess this one is tricky, but we will see others like this probably 🤔

pkg/migrations/op_drop_table.go
```diff
@@ -25,6 +25,7 @@ func (o *OpDropTable) Complete(ctx context.Context, conn db.DB, tr SQLTransforme
 }
 
 func (o *OpDropTable) Rollback(ctx context.Context, conn db.DB, tr SQLTransformer, s *schema.Schema) error {
+	s.UnRemoveTable(o.Name)
```
exekias, line 28: worth adding a comment on why this is done?

andrew-farries: comment added

andrew-farries:
> what will happen if you do a drop_table + create_table with the same name? I guess this one is tricky, but we will see others like this probably 🤔

This won't work currently. I've described the problem here: https://github.com/xataio/pgroll/issues/239#note1.

exekias:
I guess it should be fine as long as we are able to detect this situation and prevent the migration from moving forward, is that the case already?

PR #592, opened 2025-01-14
Add pg type type info to schema

### LINKED ISSUE #385: More information in `/schema` about column types.
To better support enums in the frontend, we need some more information in the `/schema` API, here is an example, this column (`Field.usage`) is an enum but we have no way of telling it from, say, a custom domain type or custom types from extensions. Eventually, we want to make all types editable from the UI. So, we need the `/schema` API to return that this is an enum and its possible values for enum's case. This will allow us to build features like dropdown select for enum and/or client side validation for enum value.

Current output:

```json
{
  "schema": {
    "name": "",
    "display_name": "",
    "tables": {
      "Field": {
        "oid": "6526469",
        "name": "Field",
        "comment": "",
        "columns": {
          "usage": {
            "name": "usage",
            "type": "bb_mmco0u98b56618sjl6glp6csh8_d5mrcj.\"FieldUsage\"",
            "default": null,
            "nullable": false,
            "unique": false,
            "comment": ""
          }
        },
        "indexes": {
          "Field_pkey": {
            "name": "Field_pkey",
            "unique": true,
            "columns": ["id"]
          }
        },
        "primaryKey": ["id"],
        "foreignKeys": {},
        "checkConstraints": {},
        "uniqueConstraints": {},
        "xataCompatible": false
      }
    }
  }
}
```

Potential future output

```json
{
  "schema": {
    "name": "",
    "display_name": "",
    "tables": {
      "Field": {
        "oid": "6526469",
        "name": "Field",
        "comment": "",
        "columns": {
          "usage": {
            "name": "usage",
            "type": "bb_mmco0u98b56618sjl6glp6csh8_d5mrcj.\"FieldUsage\"",
            "default": null,
            "nullable": false,
            "unique": false,
            "comment": "",
            "postgresType": "enum",
            "possibleValues": ["TOP_HALF", "BOTTOM_HALF", "FULL_FIELD"]
          }
        },
        "indexes": {
          "Field_pkey": {
            "name": "Field_pkey",
            "unique": true,
            "columns": ["id"]
          }
        },
        "primaryKey": ["id"],
        "foreignKeys": {},
        "checkConstraints": {},
        "uniqueConstraints": {},
        "xataCompatible": false
      }
    }
  }
}
```

## Other considerations (very hand-wavey)

1. We can add "checkConstraint" too, which has the PostgreSQL check, e.g. `{"checkConstraint": {"length(xata_id) < 256"}}`. The frontend can then potentially parse this into a TypeScript validation rule.
2. Knowing that a type is array, composite, range and their dimensions. Currently, we interpret this in the frontend with some lazy parsing, `pgroll` should be the source of truth for this information. 
3. Knowing that a type is [domain type](https://www.postgresql.org/docs/current/extend-type-system.html)
4. Knowing that a type is a [custom type](https://www.postgresql.org/docs/current/xtypes.html)
5. With all these considerations, the output of `postgresType` can be 'enum' | 'array' | 'range' | 'composite' | 'domain' | 'custom-type'

For the purpose of this ticket, `enum` is enough.

Related https://github.com/xataio/pgroll/issues/376

agedemenli:
Adding `PostgresType` into read_schema output and filling it with info obtained from `typtype` column of `pg_catalog.pg_type`. Adding a test with different type types and fixing the already existing ones.

partially fixes #385 

for reference: 
> `typtype char`
typtype is b for a base type, c for a composite type (e.g., a table's row type), d for a domain, e for an enum type, p for a pseudo-type, r for a range type, or m for a multirange type. See also typrelid and typbasetype.

from: https://www.postgresql.org/docs/current/catalog-pg-type.html#CATALOG-PG-TYPE

### DIFF
--- a/pkg/schema/schema.go
+++ b/pkg/schema/schema.go
@@ -83,6 +83,9 @@ type Column struct {
 
 	// Whether or not the column has been deleted in the virtual schema
 	Deleted bool `json:"-"`
+
+	// Postgres type type, e.g enum, composite, range
+	PostgresType string `json:"postgresType"`
 }
 
 // Index represents an index on a table
--- a/pkg/state/init.sql
+++ b/pkg/state/init.sql
@@ -163,17 +163,21 @@ BEGIN
                                         array_agg(e.enumlabel ORDER BY e.enumsortorder)
                                     FROM pg_enum AS e
                                 WHERE
-                                    e.enumtypid = tp.oid) AS enumValues,
-                                CASE 
-                                    WHEN tp.typtype='b' THEN 'base'
-                                    WHEN tp.typtype='c' THEN 'composite'
-                                    WHEN tp.typtype='d' THEN 'domain'
-                                    WHEN tp.typtype='e' THEN 'enum'
-                                    WHEN tp.typtype='p' THEN 'pseudo'
-                                    WHEN tp.typtype='r' THEN 'range'
-                                    WHEN tp.typtype='m' THEN 'multirange'
-                                END AS postgresType
-                                FROM pg_attribute AS attr
+                                    e.enumtypid = tp.oid) AS enumValues, CASE WHEN tp.typtype = 'b' THEN
+                                    'base'
+                                WHEN tp.typtype = 'c' THEN
+                                    'composite'
+                                WHEN tp.typtype = 'd' THEN
+                                    'domain'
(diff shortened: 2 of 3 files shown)


andrew-farries (approved):
This looks good 👍

PR #594, opened 2025-01-14
Add support for primary key constraints in `create_table` operation

kvch:
This PR adds support for setting primary key constraints in the `constraints` option of `create_table`.

Please note that if you set primary keys in `columns`, you are not allowed to configure `primary_key` constraints.

This configuration method is superior to marking columns as primary keys in the list `columns`, because it supports indexing options and deferring constraint checks.

Allowed options:
* `columns` (required)
* `index_parameters.tablespace`
* `index_parameters.storage_parameters`
* `index_parameters.include_columns`
* `deferrable`
* `initially_deferred`

Example:

```json
{
  "name": "50_create_table_with_table_constraint",
  "operations": [
    {
      "create_table": {
        "name": "phonebook",
        "columns": [
          {
            "name": "id",
            "type": "serial",
          },
          {
            "name": "name",
            "type": "varchar(255)"
          }
        ],
        "constraints": [
          {
            "name": "my_pk",
            "type": "primary_key",
            "columns": [
              "id"
            ] 
          }
        ]
      }
    }
  ]
}
```


pkg/sql2pgroll/create_table_test.go
```diff
@@ -247,9 +251,7 @@ func TestUnconvertableCreateTableStatements(t *testing.T) {
 		"CREATE TABLE foo(a text COLLATE en_US)",
 
 		// Table constraints, named and unnamed, are not supported
-		"CREATE TABLE foo(a int, CONSTRAINT foo_pk PRIMARY KEY (a))",
```
andrew-farries, line 250: As well as removing SQL statements for which we used to fall back to raw SQL, we need to make sure that any `CREATE TABLE` statements that we can't accurately convert to `pgroll` operations. This means any statement that has options that aren't representable by the corresponding `pgroll` operation.

It looks like from [the docs](https://www.postgresql.org/docs/current/sql-createtable.html) that all options that can be present on a `PRIMARY KEY` table constraint can be converted though. Do you know of any that can't? If so, we should add that SQL here, to make sure we don't lose information in the conversion from SQL to `pgroll`.

This comment also holds for other table constraint types that have been added so far (`unique` and `check).

andrew-farries (commented):
This looks nice 👍 

Can we also add an example migration that shows how to create PKs using a table constraint?

pkg/sql2pgroll/create_table_test.go
kvch: No, I made sure all options of primary key constraints are included:
* column list
* index parameters: setting tablespace, storage parameters and including columns
* all options for deferring checks: `deferrable`, `not deferrable`, `initially deferred`, `initially immediate`.

I also made sure to include all options for other constraint types like `unique` and `check`.

andrew-farries (approved):

Issue #582, opened 2025-01-10
Panic when attempting to backfill a table with a `uuid` primary key


ilyakooo0:
I am not familiar with Go, but from looking at the code it seems like there are restrictions on the types that primary keys can have.

It seems like they need to either be `int`s or `string`s. I have UUID primary keys and it looks like this breaks pgroll.

https://github.com/xataio/pgroll/blob/daba767a10c244cc2efcae982dc11c67fdfd61a2/pkg/migrations/backfill.go#L213-L218

<details>
<summary>trace</summary>

```
 Initialization complete
▀  Starting migration... (0s)
▀  0 records complete...
▀  1000 records complete...panic: unsupported type
goroutine 1 [running]:
github.com/xataio/pgroll/pkg/migrations.(*batchStatementBuilder).buildBatchSubQuery(0xc00033af90, {0x87e720, 0xc000333a28})
	github.com/xataio/pgroll/pkg/migrations/backfill.go:173 +0x64b
github.com/xataio/pgroll/pkg/migrations.(*batchStatementBuilder).buildQuery(0xc00033af90, {0x87e720?, 0xc000333a28?})
	github.com/xataio/pgroll/pkg/migrations/backfill.go:139 +0x37
github.com/xataio/pgroll/pkg/migrations.Backfill.(*batcher).updateBatch.func1({0x9faa40, 0xd37c20}, 0xc0000aa200)
	github.com/xataio/pgroll/pkg/migrations/backfill.go:105 +0x45
github.com/xataio/pgroll/pkg/db.(*RDB).WithRetryableTransaction(0xc0001281c0, {0x9faa40, 0xd37c20}, 0xc00004c3f0)
	github.com/xataio/pgroll/pkg/db/db.go:65 +0xe3
github.com/xataio/pgroll/pkg/migrations.(*batcher).updateBatch(...)
	github.com/xataio/pgroll/pkg/migrations/backfill.go:103
github.com/xataio/pgroll/pkg/migrations.Backfill({0x9faa40, 0xd37c20}, {0x9f9ac8, 0xc0001281c0}, 0xc0001f8af0, 0x3e8, 0x0, {0xc0001284e0, 0x1, 0x1})
	github.com/xataio/pgroll/pkg/migrations/backfill.go:41 +0x13b
[...]

PR #595, opened 2025-01-14
Simplify fetching rows with composite primary keys during backfill

kvch:
This PR simplifies the way we put together the `WHERE` clause of the query we use to fetch rows during the backfilling process. We no longer check the type of the returned identifiers, instead every literal is quoted as before.

I also added a test to check if we can backfilling works properly with composite primary keys.

Closes https://github.com/xataio/pgroll/issues/582

### DIFF
--- a/pkg/migrations/backfill.go
+++ b/pkg/migrations/backfill.go
@@ -142,6 +142,7 @@ func newBatcher(table *schema.Table, batchSize int) *batcher {
 	return &batcher{
 		statementBuilder: newBatchStatementBuilder(table.Name, getIdentityColumns(table), batchSize),
 		pkcount:          len(getIdentityColumns(table)),
+		lastValues:       make([]string, len(getIdentityColumns(table))),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/migrations/backfill.go
```diff
@@ -151,7 +151,11 @@ func (b *batcher) updateBatch(ctx context.Context, conn db.DB) error {
 
 		// Execute the query to update the next batch of rows and update the last PK
 		// value for the next batch
-		err := tx.QueryRowContext(ctx, query).Scan(&b.lastValues)
+		wrapper := make([]interface{}, len(b.lastValues))
```
ryanslade, line 154: nit: interace{} = any

```suggestion
		wrapper := make([]any, len(b.lastValues))
```

ryanslade:
LGTM. Won't approve until @andrew-farries has taken a look because auto merge is enabled.

andrew-farries (commented):
Changes look good.

#582 is about a problem with a `uuid`-valued primary key. To close it with this issue we need a test that uses such a table.

andrew-farries (approved):

PR #599, opened 2025-01-16
Add new subcommand `sql-folder` to transform SQL files into pgroll operations

kvch:
This PR adds a new subcommand called `sql-folder`.
It expects an argument, the path to the migrations folder. This folder must contain more folderes with migration files.

Example layout:

```
target-folder
  01_first_migration
    migration.sql
  02_second_migration
    migration.sql
```

The code is messy, but I put it up, so if someone else wants to test migration folders, they can use this cmd.


kvch:
Closing as it is not going to be integrated now.

PR #605, opened 2025-01-17
New option for columns: `generated`

kvch:
This PR adds support for generated columns in `create_table` and `add_column` operations.

The column definition is extended with the setting named `generated`.

Define a stored column:
```json
...
"generated": {
  "expression": "upper(name)"
}
```

Define a generated identity column:
```json
...
"generated": {
  "identity": {
    "user_specified_values": "ALWAYS",
    "sequence_options": "start 5"
}
```

Identity columns are marked not null by default, so setting `nullable` is unnecessary.

I changed the existing `fruits` table to use a generated identity column as a primary key.

```json
{
  "name": "08_create_fruits_table",
  "operations": [
    {
      "create_table": {
        "name": "fruits",
        "columns": [
          {
            "name": "id",
            "type": "bigint",
            "pk": true,
            "generated": {
              "identity": {
                "user_specified_values": "BY DEFAULT"
              }
            }
          },
          {
            "name": "name",
            "type": "varchar(255)",
            "unique": true
          },
          {
            "name": "price",
            "type": "decimal(10,2)"
          }
        ]
      }
    }
  ]
}
```

The PR also contains the transformation required by `sql2pgroll` project.

Requires https://github.com/xataio/pg_query_go/pull/7


pkg/migrations/errors.go
```diff
@@ -266,3 +266,12 @@ type PrimaryKeysAreAlreadySetError struct {
...
 	return fmt.Sprintf("table %q already has a primary key configuration in columns list", e.Table)
 }
+
+type InvalidGeneratedColumnError struct {
+	Table  string
+	Column string
+}
+
+func (e InvalidGeneratedColumnError) Error() string {
+	return fmt.Sprintf("column %q on table %q is invalid, please do not set both generated.expression and generated.identity", e.Column, e.Table)
```
andrew-farries, line 276:
```suggestion
	return fmt.Sprintf("column %q on table %q is invalid: only one of generated.expression and generated.identity may be set", e.Column, e.Table)
```

pkg/migrations/op_add_column_test.go
```diff
@@ -152,7 +152,7 @@ func TestAddColumn(t *testing.T) {
 			wantStartErr: migrations.BackfillNotPossibleError{Table: "users"},
 		},
 		{
-			name: "add serial columns",
+			name: "add serial columns with generated column",
```
andrew-farries, line 155: Should we change this test or add a new one just for generated columns?

pkg/sql2pgroll/create_table_test.go
```diff
@@ -184,13 +185,22 @@ func TestConvertCreateTableStatements(t *testing.T) {
 			sql:        "CREATE TABLE foo(b text, c text, PRIMARY KEY (b) DEFERRABLE)",
 			expectedOp: expect.CreateTableOp26,
 		},
+		{
+			sql:        "CREATE TABLE foo(b bigint GENERATED ALWAYS AS IDENTITY (START WITH 1 INCREMENT BY 1 MINVALUE 1 MAXVALUE 1000 CYCLE CACHE 1) PRIMARY KEY)",
+			expectedOp: expect.CreateTableOp27,
+		},
+		{
+			sql:        "CREATE TABLE foo(a text, b text GENERATED ALWAYS AS (upper(a)) STORED)",
```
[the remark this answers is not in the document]
kvch: We do not need to store `STORED` anywhere. When a column is generated with an expression `STORED` must be added to the end of the SQL statement, and we do add that to in `op_add_column.go`. So when `Column.Generated.Expression` is set, it is always added to the statement.

> PostgreSQL currently implements only stored generated columns.
Ref: https://www.postgresql.org/docs/current/ddl-generated-columns.html

At the moment PostgreSQL only supports stored generated columns. If they decide to add virtual columns, we can add an new attribute e.g. `Column.Generated.Virtual`.

pkg/migrations/op_add_column_test.go
kvch: I added a separate test as you requested.

pkg/sql2pgroll/create_table.go
```diff
@@ -171,11 +173,46 @@ func convertColumnDef(tableName string, col *pgq.ColumnDef) (*migrations.Column,
...
+			sequenceOptions := ""
+			if c.GetConstraint().GetOptions() != nil {
+				sequenceOptions, err = pgq.DeparseRelOptions(c.GetConstraint().GetOptions())
+				if err != nil {
+					return nil, fmt.Errorf("parsing sequence options: %w", err)
+				}
+				// transform the options into the proper format
+				sequenceOptions = strings.ReplaceAll(sequenceOptions, "=", " ")
+				sequenceOptions = strings.ReplaceAll(sequenceOptions, ",", "")
+				sequenceOptions = sequenceOptions[1 : len(sequenceOptions)-1]
```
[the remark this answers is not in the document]
kvch: The cleanest would be to expose it in `pg_query_go`. I will submit a PR soon, moving this PR into draft in the meantime.

kvch:
Converting into draft until required changes are merged in https://github.com/xataio/pg_query_go.

kvch:
`pg_query_go` PR: https://github.com/xataio/pg_query_go/pull/7

andrew-farries (approved):

Issue #485, opened 2024-11-25
Adding a `unique` constraint can fail under high load 


andrew-farries:
There is a race condition when adding a `unique` constraint to a column.

The race can cause the operation to fail, particularly when the target database is under high load at the time of the migration.

## Reproduction 1

Create a pair of migrations:

<details>
<summary><b>01_create_table.json</b></summary>

```json
{
  "name": "01_create_table",
  "operations": [
    {
      "create_table": {
        "name": "movies",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "name",
            "type": "varchar(255)"
          }
        ]
      }
    }
  ]
}
```
</details>

<details>
<summary><b>02_set_unique.json</b></summary>

```json
{
  "name": "02_set_unique",
  "operations": [
    {
      "alter_column": {
        "table": "movies",
        "column": "name",
        "unique": {
          "name": "movies_name_unique"
        },
        "up": "name",
        "down": "name"
      }
    }
  ]
}
```
</details>

Apply the first migration:

```bash
pgroll start 01_create_table.json --complete
```

In the target database, generate some load on the system:

```sql
CREATE TABLE test(a int);
INSERT INTO test(a) SELECT generate_series(1, 10000000)
```

While the above `INSERT` operation is running, apply the second migration:

```bash
pgroll start 02_set_unique.json --complete
```

The second migration is likely to fail with the following error:

```
Failed to complete migration: unable to execute complete operation: pq: index "movies_name_unique" is not valid
```

## Reproduction 2

The same problem also exists when a column with a `UNIQUE` constraint is duplicated, for example to change the nullability of the column. To reproduce:

Create a pair of migrations:

<details>
<summary><b>01_create_table.json</b></summary>

```json
{
  "name": "01_create_table",
  "operations": [
    {
      "create_table": {
        "name": "movies",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "name",
            "type": "varchar(255)",
            "unique": true,
            "nullable": true
          }
        ]
      }
    }
  ]
}
```
</details>

<details>
<summary><b>02_set_notnull.json</b></summary>

```json
{
  "name": "02_set_notnull",
  "operations": [
    {
      "alter_column": {
        "table": "movies",
        "column": "name",
        "nullable": false,
        "up": "name",
        "down": "name"
      }
    }
  ]
}
```
</details>

[...]

PR #613, opened 2025-01-21
Retry creating unique index

agedemenli:
Introduce a retry mechanism for creating unique indexes.
First, create unique index concurrently. Then wait until Postgres is done with the index creation, periodically checking the view `pg_stat_progress_create_index`. Once it is done, lookup pg_index to see if the index is validated or not. If it is, we are good to go. If not, drop the index and try again.

Since we are running select queries on `pg_stat_progress_create_index` and `pg_index` and we are expecting a real output to see the progress, this approach doesn't work with fake db. I have added hardcoded responses for the fake db scenario, so that we can safely process migrations to update virtual schema as well.
Not sure if this is a good solution or not. Open for discussions.

Tested manually with high load, while 10m rows being inserted. But I wonder if there's a way to add a test for the "high load" scenario.

### DIFF
--- a/pkg/migrations/duplicate.go
+++ b/pkg/migrations/duplicate.go
@@ -40,7 +40,6 @@ const (
 	dataTypeMismatchErrorCode  pq.ErrorCode = "42804"
 	undefinedFunctionErrorCode pq.ErrorCode = "42883"
 
-	cCreateUniqueIndexSQL            = `CREATE UNIQUE INDEX CONCURRENTLY %s ON %s (%s)`
 	cSetDefaultSQL                   = `ALTER TABLE %s ALTER COLUMN %s SET DEFAULT %s`
 	cAlterTableAddCheckConstraintSQL = `ALTER TABLE %s ADD CONSTRAINT %s %s NOT VALID`
 	cAlterTableAddForeignKeySQL      = `ALTER TABLE %s ADD CONSTRAINT %s FOREIGN KEY (%s) REFERENCES %s (%s) ON DELETE %s`
@@ -134,12 +133,16 @@ func (d *Duplicator) Duplicate(ctx context.Context) error {
 		}
 	}
 
-	// Generate SQL to duplicate any unique constraints on the columns
-	// The constraint is duplicated by adding a unique index on the column concurrently.
+	// Create indexes for unique constraints on the columns concurrently.
 	// The index is converted into a unique constraint on migration completion.
-	for _, sql := range d.stmtBuilder.duplicateUniqueConstraints(d.withoutConstraint, colNames...) {
-		if _, err := d.conn.ExecContext(ctx, sql); err != nil {
-			return err
+	for _, uc := range d.stmtBuilder.table.UniqueConstraints {
+		if slices.Contains(d.withoutConstraint, uc.Name) {
+			continue
+		}
+		if duplicatedMember, constraintColumns := d.stmtBuilder.allConstraintColumns(uc.Columns, colNames...); duplicatedMember {
+			if err := createUniqueIndexConcurrently(ctx, d.conn, "", DuplicationName(uc.Name), d.stmtBuilder.table.Name, constraintColumns); err != nil {
+				return err
+			}
 		}
 	}
 
@@ -174,23 +177,6 @@ func (d *duplicatorStmtBuilder) duplicateCheckConstraints(withoutConstraint []st
 				pq.QuoteIdentifier(d.table.Name),
 				pq.QuoteIdentifier(DuplicationName(cc.Name)),
 				rewriteCheckExpression(cc.Definition, duplicatedConstraintColumns...),
-			))
-		}
-	}
-	return stmts
-}
-
-func (d *duplicatorStmtBuilder) duplicateUniqueConstraints(withoutConstraint []string, colNames ...string) []string {
-	stmts := make([]string, 0, len(d.table.UniqueConstraints))
-	for _, uc := range d.table.UniqueConstraints {
-		if slices.Contains(withoutConstraint, uc.Name) {
-			continue
-		}
-		if duplicatedMember, constraintColumns := d.allConstraintColumns(uc.Columns, colNames...); duplicatedMember {
-			stmts = append(stmts, fmt.Sprintf(cCreateUniqueIndexSQL,
-				pq.QuoteIdentifier(DuplicationName(uc.Name)),
-				pq.QuoteIdentifier(d.table.Name),
-				strings.Join(quoteColumnNames(constraintColumns), ", "),
 			))
 		}
 	}
--- a/pkg/migrations/duplicate_test.go
+++ b/pkg/migrations/duplicate_test.go
@@ -228,7 +228,7 @@ func TestCreateIndexConcurrentlySqlGeneration(t *testing.T) {
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
-			stmt := getCreateUniqueIndexConcurrentlySql(testCases.indexName, testCases.schemaName, testCases.tableName, testCases.columns)
+			stmt := getCreateUniqueIndexConcurrentlySQL(testCases.indexName, testCases.schemaName, testCases.tableName, testCases.columns)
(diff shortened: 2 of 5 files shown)


pkg/migrations/op_set_unique.go
```diff
@@ -83,3 +115,48 @@ func addUniqueIndex(ctx context.Context, conn db.DB, table, column, name string)
 
 	return err
 }
+
+func isIndexInProgress(ctx context.Context, conn db.DB, schemaname string, indexname string) (bool, error) {
+	var isInProgress bool
```
kvch, line 120: Nit: please move `inProgress` to line 135 (before using it in `ScanFirstValue).

pkg/migrations/op_set_unique.go
```diff
@@ -25,12 +26,43 @@ func (o *OpSetUnique) Start(ctx context.Context, conn db.DB, latestSchema string
 	table := s.GetTable(o.Table)
 	column := table.GetColumn(o.Column)
 
-	// Add a unique index to the new column
-	if err := addUniqueIndex(ctx, conn, table.Name, column.Name, o.Name); err != nil {
-		return nil, fmt.Errorf("failed to add unique index: %w", err)
+	for retryCount := 5; retryCount > 0; retryCount-- {
```
kvch, line 29: This format is more idiomatic:

```suggestion
	for retryCount := 0; retryCount < 5; retryCount++ {
```

pkg/migrations/op_set_unique.go
```diff
@@ -25,12 +26,43 @@ func (o *OpSetUnique) Start(ctx context.Context, conn db.DB, latestSchema string
...
+		isInProgress := true
+		for isInProgress {
+			var err error
+			isInProgress, err = isIndexInProgress(ctx, conn, s.Name, o.Name)
+			if err != nil {
+				return nil, err
+			}
+			// Still in progress, sleep for 0.5 seconds and check again
+			time.Sleep(500 * time.Millisecond)
+		}
```
andrew-farries, line 45: This is always going to sleep for 500ms, regardless of whether the index creation succeeds on the first iteration. Maybe we can rewrite to avoid the sleep if `isInProgress` is false?

pkg/migrations/op_set_unique.go
```diff
@@ -83,3 +115,48 @@ func addUniqueIndex(ctx context.Context, conn db.DB, table, column, name string)
...
+		SELECT EXISTS(
+			SELECT * FROM pg_catalog.pg_stat_progress_create_index
+			WHERE index_relid = $1::regclass
+			)`,
+		fmt.Sprintf("%s.%s", pq.QuoteIdentifier(schemaname), pq.QuoteIdentifier(indexname)))
+	if err != nil {
+		return false, fmt.Errorf("getting index in progress with name %q: %w", indexname, err)
+	}
+	if rows == nil {
+		// if rows == nil && err != nil, then it means we have queried a fake db.
```
andrew-farries, line 131:
```suggestion
		// if rows == nil && err != nil, then it means we have queried a `FakeDB`.
```

Make it clear we're talking about the `FakeDB` implementation of our `DB` interface.

andrew-farries (commented):
This looks good 👍 

There are other places where we create indexes concurrently that might fail in a similar way; https://github.com/xataio/pgroll/issues/485 lists another (under **Reproduction 2**), so we might want to extract this to something that we can re-use for these other cases too in later PRs

pkg/migrations/op_set_unique.go
```diff
@@ -25,12 +26,43 @@ func (o *OpSetUnique) Start(ctx context.Context, conn db.DB, latestSchema string
...
-	if err := addUniqueIndex(ctx, conn, table.Name, column.Name, o.Name); err != nil {
-		return nil, fmt.Errorf("failed to add unique index: %w", err)
+	for retryCount := 5; retryCount > 0; retryCount-- {
+		// Add a unique index to the new column
+		if err := addUniqueIndex(ctx, conn, table.Name, column.Name, o.Name); err != nil {
+			return nil, fmt.Errorf("failed to add unique index: %w", err)
+		}
+
+		// Make sure Postgres is done creating the index
+		isInProgress := true
```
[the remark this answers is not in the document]
agedemenli: done, thanks!

pkg/migrations/index.go
```diff
@@ -0,0 +1,125 @@
...
+
+		ticker := time.NewTicker(500 * time.Millisecond)
+		for isInProgress {
+			<-ticker.C
+			isInProgress, err = isIndexInProgress(ctx, conn, quotedQualifiedIndexName)
+			if err != nil {
+				return err
+			}
+		}
+		ticker.Stop()
```
kvch, line 41: Please move `ticker.Stop` after the line where you initialize it, and add defer:
```golang
ticker := time.NewTicker(500 * time.Millisecond)
defer ticker.Stop()
```

With this change we make sure that if `isIndexInProgress` returns an error, the `ticker` is stopped.

pkg/migrations/duplicate.go
```diff
@@ -134,12 +133,16 @@ func (d *Duplicator) Duplicate(ctx context.Context) error {
...
 	// The index is converted into a unique constraint on migration completion.
-	for _, sql := range d.stmtBuilder.duplicateUniqueConstraints(d.withoutConstraint, colNames...) {
-		if _, err := d.conn.ExecContext(ctx, sql); err != nil {
-			return err
+	for _, uc := range d.stmtBuilder.table.UniqueConstraints {
+		if slices.Contains(d.withoutConstraint, uc.Name) {
+			continue
+		}
+		if duplicatedMember, constraintColumns := d.stmtBuilder.allConstraintColumns(uc.Columns, colNames...); duplicatedMember {
+			if err := createUniqueIndexConcurrently(ctx, d.conn, "", DuplicationName(uc.Name), d.stmtBuilder.table.Name, constraintColumns); err != nil {
```
kvch, line 143: Nice

andrew-farries, line 143: Is passing `""` as the schema name here safe? 

What happens if we are duplicating a unique constraint on a column on a table in a non-`public` schema?

pkg/migrations/duplicate_test.go
```diff
@@ -102,47 +102,6 @@ func TestDuplicateStmtBuilderCheckConstraints(t *testing.T) {
 	}
 }
 
-func TestDuplicateStmtBuilderUniqueConstraints(t *testing.T) {
```
andrew-farries, line 105: Is it possible to move these tests rather than remove them? Or test the same SQL generation some other way?

pkg/migrations/duplicate.go
agedemenli: It is as safe as it was before this PR. We were already generating create index statements using `pq.QuoteIdentifier(d.table.Name)`, without schema name. No change in behavior. 

Though, I'd prefer to always have schema name, using qualified names instead of just object names. I was not able to do it here because at this point of the code we don't have it. We should maybe consider adding schema name info into `Duplicator` or `duplicatorStmtBuilder`.

pkg/migrations/duplicate_test.go
agedemenli: Because the SQL generation function will not be used anymore and also is removed by this PR, I see no point in keeping this test. Maybe we can add another test for SQL statements generated in `executeCreateUniqueIndexConcurrently` if it makes sense?

pkg/migrations/duplicate.go
agedemenli: https://github.com/xataio/pgroll/blob/main/pkg/migrations/duplicate.go#L208

For instance, we don't use schema name here as well. What happens if the table is in a non-public schema?

andrew-farries: Ok, it works because `pgroll` sets the Postgres `search_path` to whatever the `--schema` flag is set to:

https://github.com/xataio/pgroll/blob/c09619aa879e0a0e8782eb495962cfe8a800a559/pkg/roll/roll.go#L100-L101

So there is no need to explicitly schema qualify names here at all.

andrew-farries (approved):

kvch (approved):

PR #624, opened 2025-01-24
Add support for exclusion constraints in `create_table` operation

kvch:
This PR adds a new type of constraint to `create_table` operation called `exclude` to support [exclusion constraints](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-EXCLUSION).

New options:
* `exclude.index_method`: name of the index method (e.g. btree, gist, etc.)
* `exclude.elements`: exlusion elements
* `exclude.predicate`: WHERE clause for partial indices

Example constraint forbidding overlap in room reservations:
```json
{
  "name": "52_create_table_with_table_exclusion_constraint",
  "operations": [
    {
      "create_table": {
        "name": "room_reservations",
        "columns": [
          {
            "name": "id",
            "type": "serial"
          },
          {
            "name": "start",
            "type": "timestamp"
          },
          {
            "name": "end",
            "type": "timestamp"
          },
          {
            "name": "canceled",
            "type": "boolean",
            "default": "false"
          }
        ],
        "constraints": [
          {
            "name": "forbid_double_booking",
            "type": "exclude",
            "exclude": {
              "index_method": "gist",
              "elements": "id WITH =, tsrange(\"start\", \"end\") WITH &&",
              "predicate": "NOT canceled"
            }
          }
        ]
      }
    }
  ]
}
```

The tests I added require the extension `btree_gist`. So I added it for `make examples` and running `go test`.


go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.35.0
 	github.com/testcontainers/testcontainers-go/modules/postgres v0.35.0
-	github.com/xataio/pg_query_go/v6 v6.0.0-20250122133641-54118c062181
+	github.com/xataio/pg_query_go/v6 v6.0.0-20250123182324-526a22cbe0c0
```
kvch, line 17: This will be updated when my change is merge in `pg_query_go`.

.github/workflows/build.yml
```diff
@@ -220,6 +220,7 @@ jobs:
         if [ "$PGROLL_SCHEMA" != "public" ]; then
           psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -c "CREATE SCHEMA $PGROLL_SCHEMA;"
         fi
+        psql -h $POSTGRES_HOST -p $POSTGRES_PORT -U postgres -c "CREATE EXTENSION IF NOT EXISTS btree_gist SCHEMA $PGROLL_SCHEMA;"
```
andrew-farries, line 223: We need to make a similar change to the local `docker compose` setup too:

```
docker compose up
make examples
```

now fails because the `btree_gist` extension isn't installed.

examples/52_create_table_with_exclusion_constraint.json
```diff
@@ -0,0 +1,56 @@
+{
```
andrew-farries, line 1: Is it possible to give an example migration that doesn't require the `btree_gist` extension?

All the example migrations so far run against a vanilla Postgres - this is the first one that requires an extension.

pkg/migrations/op_create_table_test.go
```diff
@@ -1020,6 +1020,191 @@ func TestCreateTable(t *testing.T) {
 				}, rows)
 			},
 		},
+		{
```
andrew-farries, line 1023: We could potentially just keep the first of these tests, which wouldn't require the extension.

Do you think that still provides enough coverage?

andrew-farries (commented):
Looks good 👍 

I wonder whether we can do without the need for the `btree_gist` extension though.

kvch:
I think the simple test provides enough coverage already. I am happy to remove the more complex tests. \o/

andrew-farries (approved):

PR #628, opened 2025-01-27
Unify table level foreign key conversion in `sql2pgroll`

kvch:
This PR unifies FK constraints in `sql2pgroll` package. The new function `convertFkConstraint` in is the unified way to process table level FK constraints of columns and tables.


pkg/sql2pgroll/alter_table.go
```diff
@@ -185,83 +185,27 @@ func convertAlterTableAddForeignKeyConstraint(stmt *pgq.AlterTableStmt, constrai
...
-	switch constraint.GetFkMatchtype() {
-	case "f":
-		// FULL
-		return false
-	case "s":
-		// SIMPLE, the default
-		break
-	}
-	return true
+	return !constraint.SkipValidation
```
kvch, line 208: The constraint writer already supports skipping validation.

schema.json
```diff
@@ -113,20 +113,86 @@
...
+        },
         "on_delete": {
           "description": "On delete behavior of the foreign key constraint",
-          "$ref": "#/$defs/ForeignKeyReferenceOnDelete",
+          "$ref": "#/$defs/ForeignKeyOnDelete",
+          "default": "NO ACTION"
+        },
+        "on_update": {
+          "description": "On update behavior of the foreign key constraint",
+          "$ref": "#/$defs/ForeignKeyOnDelete",
```
andrew-farries, line 128: Looks like the wrong `$ref` here.

andrew-farries (commented):
I think this looks good 👍 

It would be good to separate out the simpler changes (eg constant renames) from other changes so that individual PRs don't get too big.

schema.json
kvch: No, it is on purpose. `ON UPDATE` and `ON DELETE` have the same actions, so we can reuse it in both places. But I can rename it to make clearer that it is intentional.

kvch:
Requires https://github.com/xataio/pgroll/pull/639.
Until that PR is merged, it can stay open as a draft.

PR #629, opened 2025-01-27
Add script to generate CLI structure

SferaDev:
Add script to generate the structure of the CLI for calling it programmatically and validating params. It will also be useful to generate documentation for pgroll.com.


cmd/root.go
```diff
@@ -16,33 +16,6 @@ import (
 // Version is the pgroll version
 var Version = "development"
 
-func init() {
```
SferaDev, line 19: I checked and it's not mandatory to have the `init()` function accoring to some GitHub discussions in the spf13/cobra repo. Added a way that we can init the rootCmd without calling `execute` to introspect the commands.

andrew-farries (commented):
This looks great; thanks for doing it 🙏 

To merge I think we also need:

* A `Makefile` rule to generate `cli-definition.json`
* Invoke the `Makefile` rule from the CI and fail the build if `cli-definition.json` is out of sync.

cmd/sql.go
```diff
@@ -13,10 +13,11 @@ import (
 
 func sqlCmd() *cobra.Command {
 	sqlCmd := &cobra.Command{
-		Use:    "sql <sql statement>",
-		Short:  "Convert SQL statements to pgroll operations",
-		Args:   cobra.ExactArgs(1),
-		Hidden: true,
+		Use:       "sql <sql statement>",
```
SferaDev, line 16:
```suggestion
		Use:       "sql <statement>",
```

SferaDev:
@andrew-farries Added it to the `generate` rule, let me know if that works or we should have a different one for this

andrew-farries:
> @andrew-farries Added it to the `generate` rule, let me know if that works or we should have a different one for this

That works for me.

Issue #583, opened 2025-01-10
`Backfill` process never completes during periods of high `INSERT` query rates


santhosh-programmer:
Hello team,

For testing, I tried this example [14_add_reviews_table.json](https://github.com/xataio/pgroll/blob/main/examples/14_add_reviews_table.json) to create a `reviews` table and then [17_add_rating_column.json](https://github.com/xataio/pgroll/blob/main/examples/17_add_rating_column.json) to add a `rating` column. 

I have inserted 50k dummy rows initially to this table.

Now to test `alter column` operation, i started [18_change_column_type.json](https://github.com/xataio/pgroll/blob/main/examples/18_change_column_type.json) migration and  i ran `pgbench`  parallelly  as follows,

```
pgbench -f insert_script.sql -c 1 -T 2147483647 postgres 
```
<details>
<summary><b> insert_script.sql</b></summary>

```sql
INSERT INTO public.reviews (username, product, review, rating)
SELECT 
    'user_' || generate_series(1, 10000), 
    'Product_' || generate_series(1, 10000), 
    'This is a dummy review for Product_' || generate_series(1, 10000),
    (random() * 5)::integer::text;
```
</details>

Since the table is loaded with high `INSERT`  query rates from `pgbench`, the backfill procedure is not ending. Though initially 50k rows was present in the table, the backfilling continued for all the newer rows inserted by pgbench. 

Since the `trigger` is created before starting backfill, i think it is redundant to backfill the newer rows which are inserted after `trigger` creation as backfill for this rows will be taken care by the `trigger` created earlier.  Below is the console screenshot where the backfilling continued for 6 million+ rows created with `pgbench` and running infinitely.

![image](https://github.com/user-attachments/assets/ef4c2919-cc9d-4cdd-a24a-bb2f929d61a8)

In a large scale production environment, we can expect high volume of `INSERT` queries because multiple connections parallelly sends the `INSERT` queries. In such case, what should be done to avoid infinite backfilling problem? Should we block `INSERT` queries by manually acquiring some `LOCK` until backfill is completed ? If so, does it violates the concept of zero-downtime ?

andrew-farries:
Thanks for opening this @santhosh-programmer 🙏 

I've added the issue to the v1 milestone as it's something we need to address.

We need some way to prevent the backfill process from touching new rows added after backfill start. A couple of ideas OTOH:

* Use transaction ids to ensure that only rows visible before backfill start are considered for each batch
[...]

PR #632, opened 2025-01-28
Use `text/template` for backfill query generation

andrew-farries:
Generate the backfill query with a `text/template` instead of the `batchStatementBuilder`.

Using a `text/template` instead of the builder with its `Sprintf` statements make the structure of the query easier to see and modify as part of #583.

Part of #583


kvch (commented):
Changes look good. Do you intend to adopt `text/template` package in all SQL generation parts of pgroll?

kvch (approved):

andrew-farries:
> Changes look good. Do you intend to adopt `text/template` package in all SQL generation parts of pgroll?

I don't think it's worth doing everywhere, only for generation of longer SQL statements or functions. The only other place where it's currently used is for trigger and trigger function generation:

https://github.com/xataio/pgroll/blob/c2370016520255ce6b4c5618e2d46ae47edb79ab/pkg/migrations/templates/function.go#L5-L32

The short SQL statements that are generated elsewhere wouldn't benefit from the extra readability that using `text/template` brings.

PR #634, opened 2025-01-29

andrew-farries:
Implement one possible solution to limit backfill to touch only rows that existed in the table prior to backfill start and ignore rows inserted or updated after backfill start.

I believe the solution presented here is flawed and should not be merged as is; the PR is open for discussion to see if the technique can be made to work or if we need to take a different approach.

In general, it is fine from a correctness POV if backfill updates a row that was already updated/inserted by a transaction that committed after backfill started; it's not OK from a performance POV however as a backfill running at the same time as a high rate of `INSERT`s into the table will cause the backfill to never terminate (the issue is described in #583).

## Proposed solution

Backfill works be touching rows in batches (batch size is configurable). 'Touch' in this context means to set the row's PK to itself, causing the already-installed backfill trigger to fire for that row.

As of this PR, the per-batch query looks like:

```sql
WITH batch AS
(
  SELECT "id", "zip"
  FROM "table_name"
  WHERE (
    "pgroll".b_follows_a(xmin::text::bigint, 1234) OR
    "pgroll".b_follows_a(xmin::text::bigint, "pgroll".frozen_xid('public', 'table_name')::text::bigint)
  )
  AND ("id", "zip") > ('1', '1234')
  ORDER BY "id", "zip"
  LIMIT 10
  FOR NO KEY UPDATE
),
update AS
(
  UPDATE "table_name"
  SET "id" = "table_name"."id", "zip" = "table_name"."zip"
  FROM batch
  WHERE "table_name"."id" = batch."id" AND "table_name"."zip" = batch."zip"
  RETURNING "table_name"."id", "table_name"."zip"
)
SELECT LAST_VALUE("id") OVER(), LAST_VALUE("zip") OVER()
FROM update
```

The first CTE (`WITH batch AS...`) is the relevant part here. The purpose of this CTE is to select the next batch of rows to be updated (and lock those rows for update).

The relevant part of the first CTE is this bit:

```sql
WHERE (
    "pgroll".b_follows_a(xmin::text::bigint, 1234) OR
    "pgroll".b_follows_a(xmin::text::bigint, "pgroll".frozen_xid('public', 'table_name')::text::bigint)
)
```

This is where we attempt to filter out tuples that were created/updated after the backfill process started. '1234' represents the `xid` of when the backfill process started. The first part:

```sql
    "pgroll".b_follows_a(xmin::text::bigint, 1234)
```

checks to see if the tuple is older than the `xid` when the backfill started. If so the tuple should be part of the batch. `b_follows_a` implements an `xid`-wraparound safe comparison of `xid`s. The transaction id space (0 - 2^32-1) is considered as a circle and anything in the forward half of the circle is considered ahead of `xid`, anything else is behind. See [Postgres Internals book - Chapter 7, Freezing](https://edu.postgrespro.com/postgresql_internals-14_en.pdf) for a description:

image

Using this calculation alone to determine relative ages between transaction ids will fail for very old rows (older than 2^31 transactions since backfill start), which will appear to be in the future. 

<div align="center">
image
</div>
Here, the first snowflake transaction at the top of the circle will appear to be in the future relative to T1, but really it's in the past. Postgres solves this problem by 'freezing` old tuples once they are guaranteed visible to all transactions (exactly when Postgres freezes old tuples is configurable, but not relevant here). So we also need a second check to always include a row in the batch if it's frozen:

```sql
    "pgroll".b_follows_a(xmin::text::bigint, "pgroll".frozen_xid('public', 'table_name')::text::bigint)
```

The `frozen_xid` function is defined:

```sql
-- Find the xid for a table before which all transaction ids in the table are
-- frozen
CREATE OR REPLACE FUNCTION pgroll.frozen_xid(schema_name name, table_name name)
RETURNS xid
LANGUAGE SQL
AS $$
  SELECT relfrozenxid
	FROM pg_class
	WHERE relnamespace::regnamespace::name = schema_name
	AND relname = table_name
	AND relkind = 'r'
$$;
```

The test checks if the transaction id that created the tuple comes before the oldest unfrozen tuple in the table  (`pg_class.relfrozenxid`). If so, the tuple is frozen and should be included in the batch even if the visibility check would regard it as in the future.

## Problem

What happens if a tuple was frozen many billions of transactions ago (ie several `xid` wraparounds ago)? the 32 bit `pg_class.relfrozenxid` won't be able to tell us accurately whether the `xid` of this extremely old tuple should be considered frozen or not - `relfrozenxid` doesn't contain `epoch` information about which wraparound cycle it refers to.

The ultimate truth of a whether a tuple is frozen or not is contained in the tuple header - frozen tuples have their `HEAP_XMIN_FROZEN` bits set in `t_infomask`:

```c
static inline bool
HeapTupleHeaderXminFrozen(const HeapTupleHeaderData *tup)
{
	return (tup->t_infomask & HEAP_XMIN_FROZEN) == HEAP_XMIN_FROZEN;
}
```

But the only way to access the tuple header is via the `pageinspect` extension. Prior to Postges `9.4`, frozen tuples had their `xmin` replaced with a special value to indicate that the row was frozen which made easy identification of frozen tuples from SQL possible, but this is no longer the case - the tuple `infomask` is used instead.

## Summary

Without a reliable way to check from SQL whether a tuple is frozen, I don't think this approach is robust. Reliably checking whether a tuple is frozen looks like it requires access to the tuple header, not possible from SQL, only by using extensions.

Using `pageinspect` to determine if a tuple is frozen may be possible, but would introduce a dependency on that extension; currently `pgroll` does not require any extensions. 

Without robust checks for frozen tuples, the backfill process could exclude tuples that should be backfilled potentially resulting in data loss.

## References
* [Postgres Internals book - Chapter 7, Freezing](https://edu.postgrespro.com/postgresql_internals-14_en.pdf)
* [Postgres Wiki article on transaction wraparound and tuple freezing](https://en.wikibooks.org/wiki/PostgreSQL/Wraparound_and_Freeze)
* [Postgres docs on vacuuming and xid wraparound](https://www.postgresql.org/docs/current/routine-vacuuming.html)
* [HOWTO article on xid wraparound](https://gitlab.com/postgres-ai/postgresql-consulting/postgres-howtos/-/blob/main/0044_how_to_monitor_transaction_id_wraparound_risks.md?ref_type=heads)
* [Short thread on Postgres hackers list about a very similar issue](https://www.postgresql.org/message-id/CAA6U2pZx7%3D2YS9u_ZZYSiKR12tbPS81hh37SqEe8t3L_WSvrUw%40mail.gmail.com)
* [SO question and answer, closely related to this issue](https://stackoverflow.com/questions/69316933/how-posgresql-marks-frozen-rows)

### DIFF
--- a/pkg/backfill/backfill.go
+++ b/pkg/backfill/backfill.go
@@ -56,12 +56,18 @@ func (bf *Backfill) Start(ctx context.Context, table *schema.Table) error {
 		return fmt.Errorf("get row count for %q: %w", table.Name, err)
 	}
 
+	xid, err := getTransactionID(ctx, bf.conn)
+	if err != nil {
+		return fmt.Errorf("get transaction id: %w", err)
+	}
+
 	// Create a batcher for the table.
 	b := batcher{
 		BatchConfig: templates.BatchConfig{
-			TableName:  table.Name,
-			PrimaryKey: identityColumns,
-			BatchSize:  bf.batchSize,
+			TableName:     table.Name,
+			PrimaryKey:    identityColumns,
+			BatchSize:     bf.batchSize,
+			TransactionID: xid,
 		},
 	}
 
@@ -86,6 +92,24 @@ func (bf *Backfill) Start(ctx context.Context, table *schema.Table) error {
 	}
 
 	return nil
+}
+
+// getTransactionID returns the current transaction ID. The transaction ID is
+// used to ensure that the backfill operation ignores rows that were
+// inserted/updated after the backfill process started.
+func getTransactionID(ctx context.Context, conn db.DB) (int64, error) {
+	var txid int64
+
+	rows, err := conn.QueryContext(ctx, "SELECT txid_current()")
+	if err != nil {
+		return 0, fmt.Errorf("getting transaction id: %w", err)
+	}
+	defer rows.Close()
+
+	if err := db.ScanFirstValue(rows, &txid); err != nil {
+		return 0, fmt.Errorf("scanning transaction id: %w", err)
+	}
+	return txid, nil
 }
 
 // getRowCount will attempt to get the row count for the given table. It first attempts to get an
--- a/pkg/backfill/backfill.go
+++ b/pkg/backfill/backfill.go
@@ -71,6 +71,7 @@ func (bf *Backfill) Start(ctx context.Context, table *schema.Table) error {
 			PrimaryKey:    identityColumns,
 			BatchSize:     bf.batchSize,
 			TransactionID: xid,
+			Schema:        bf.schema,
 			StateSchema:   bf.stateSchema,
 		},
 	}
--- a/pkg/backfill/backfill.go
+++ b/pkg/backfill/backfill.go
@@ -15,10 +15,11 @@ import (
 )
 
 type Backfill struct {
-	conn       db.DB
-	batchSize  int
-	batchDelay time.Duration
-	callbacks  []CallbackFn
+	conn        db.DB
+	batchSize   int
+	batchDelay  time.Duration
+	callbacks   []CallbackFn
+	stateSchema string
 }
 
 type CallbackFn func(done int64, total int64)
@@ -27,8 +28,9 @@ type CallbackFn func(done int64, total int64)
 // not started until `Start` is invoked.
 func New(conn db.DB, opts ...OptionFn) *Backfill {
 	b := &Backfill{
-		conn:      conn,
-		batchSize: 1000,
+		conn:        conn,
+		batchSize:   1000,
+		stateSchema: "pgroll",
 	}
 
 	for _, opt := range opts {
@@ -68,6 +70,7 @@ func (bf *Backfill) Start(ctx context.Context, table *schema.Table) error {
 			PrimaryKey:    identityColumns,
 			BatchSize:     bf.batchSize,
 			TransactionID: xid,
+			StateSchema:   bf.stateSchema,
 		},
 	}
 
--- a/pkg/backfill/backfill.go
+++ b/pkg/backfill/backfill.go
@@ -19,6 +19,7 @@ type Backfill struct {
 	batchSize   int
 	batchDelay  time.Duration
 	callbacks   []CallbackFn
+	schema      string
 	stateSchema string
 }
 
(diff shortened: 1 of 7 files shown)


andrew-farries:
I tested this solution with the aid of the `xid_wraparound` extension from the Postgres source tree [[code](https://github.com/postgres/postgres/tree/master/src/test/modules/xid_wraparound), [blog](https://masahikosawada.github.io/en/2023/12/21/Burning-XIDs/)] to be able to burn through transaction ids quickly.

Created a Postgres image with the extension installed:

```Dockerfile
FROM gcc:latest AS builder

# Clone the Postgres 17 release branch at depth 1.
# The Postgres version we checkout here must match the version of Postgres run
# in the final image, otherwise there will be a mismatch between the extension
# and the server versions.
WORKDIR /tmp
RUN git clone --branch REL_17_2 --depth=1 https://github.com/postgres/postgres.git

# Install missing build-time dependencies for Postgres
RUN apt update && apt install -y \
  make \
  flex \
  bison

# Run the 'configure' script
WORKDIR postgres
RUN ./configure

# Build the 'xid_wraparound' extension
WORKDIR src/test/modules/xid_wraparound
RUN make

# Build the final image
FROM postgres:17

# Install the extension files from the builder image
RUN mkdir -p /usr/lib/postgresql/17/modules
COPY --from=builder /tmp/postgres/src/test/modules/xid_wraparound/xid_wraparound.so /usr/lib/postgresql/17/lib/
COPY --from=builder /tmp/postgres/src/test/modules/xid_wraparound/xid_wraparound--*.sql /usr/share/postgresql/17/extension/
COPY --from=builder /tmp/postgres/src/test/modules/xid_wraparound/xid_wraparound.control /usr/share/postgresql/17/extension/

# init.sql ensures that the extension is loaded when the container is started
COPY init.sql /docker-entrypoint-initdb.d/
```

`init.sql` is:

```
CREATE EXTENSION xid_wraparound;
```

With this it's possible to approach transaction id wraparound quickly by using xids 1 billion at a time:

```sql
select consume_xids(1_000_000_000);
vacuum 
```

As expected, once transaction ids surpass 2^32-1, the `pg_class.relfrozenxid` for the table I was testing against did not wrap around (it's a 64 bit integer). Without epoch counters or a 64 bit `xmin` in the tuple header we therefore cant compare `pg_class.relfrozenxid` for the table with the `xmin` values for each to determine whether a tuple is frozen or not.

andrew-farries:
Closing in favor of #652

PR #637, opened 2025-01-29
Forbid adding generated columns to existing tables

kvch:
This is a temporary limitation we have to introduce to make sure
people are not adding generated columns to their existing tables.
Right now, with the current approach if someone adds a new generated
column to their big table, PostgreSQL has to write the new stored
value or generate the new identity for every row, leading to possible downtime.


kvch:
Improvement tracked in: https://github.com/xataio/pgroll/issues/638

andrew-farries (commented):
Can we remove the tests and add them to #638 rather than commenting out?

andrew-farries (approved):

PR #640, opened 2025-01-29
Reuse table level foreign key references in `create_constraint`

kvch:
This PR reuses the same object from constraint definition of `create_table`.
Also, match type gets its dedicated definition, as it can be reused in multiple
places as well.

Extracted from https://github.com/xataio/pgroll/pull/628


andrew-farries (commented):
Looks good. Are there any updates to the documentation that need to go along with this?

kvch:
No, this change does not have any change in the way users interact with foreign keys. I just split out the renaming as you requested in #628 to reduce the changeset.

andrew-farries (approved):

Issue #636, opened 2025-01-29
No backfill needed for empty table


agedemenli:
Primary key or a `UNIQUE NOT NULL` column is required for backfills. When running DDL on empty (probably newly created) tables, `pgroll` errors out because the backfill fails. Since backfill is actually not needed for an empty table, we should consider skipping the backfill step for empty tables. That would allow users to modify the tables that they just created freely.

PR #642, opened 2025-01-30
Skip backfill for empty table

agedemenli:
No need to backfill an empty table. Backfill fails for tables without identity columns, but it shouldn't be necessary for empty tables. This causes meaningless errors like in the example given in the issue description.

Fixes: #636

### DIFF
--- a/pkg/migrations/op_add_column_test.go
+++ b/pkg/migrations/op_add_column_test.go
@@ -112,6 +112,65 @@ func TestAddColumn(t *testing.T) {
 			},
 		},
 		{
+			name: "a newly added column can't be used as the identity column for a backfill",
+			migrations: []migrations.Migration{
+				{
+					Name: "01_add_table",
+					Operations: migrations.Operations{
+						&migrations.OpCreateTable{
+							Name: "users",
+							Columns: []migrations.Column{
+								{
+									Name: "name",
+									Type: "varchar(255)",
+								},
+							},
+						},
+					},
+				},
+				{
+					// This column is NOT NULL and UNIQUE and is added to a table without
+					// a PK, so it could in theory be used as the identity column for a
+					// backfill. However, it shouldn't be used as the identity column for
+					// a backfill because it's a newly added column whose temporary
+					// `_pgroll_new_description` column will be full of NULLs for any
+					// existing rows in the table.
+					Name: "02_add_column",
+					Operations: migrations.Operations{
+						&migrations.OpAddColumn{
+							Table: "users",
+							Column: migrations.Column{
+								Name:     "description",
+								Type:     "integer",
+								Nullable: false,
+								Unique:   true,
+							},
+							Up: "'this is a description'",
+						},
+					},
+				},
+			},
+			afterRollback: func(t *testing.T, db *sql.DB, schema string) {
+				// The new column has been dropped from the underlying table
+				columnName := migrations.TemporaryName("age")
+				ColumnMustNotExist(t, db, schema, "users", columnName)
+
+				// The table's column count reflects the drop of the new column
+				TableMustHaveColumnCount(t, db, schema, "users", 1)
+
+				// Insert a row to make sure we get a not possible error because there's no identity column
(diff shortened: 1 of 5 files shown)


pkg/migrations/op_add_column_test.go
```diff
@@ -255,6 +273,58 @@ func TestAddColumn(t *testing.T) {
...
+			},
+			afterRollback: func(t *testing.T, db *sql.DB, schema string) {
+				// The new column has been dropped from the underlying table
+				columnName := migrations.TemporaryName("age")
+				ColumnMustNotExist(t, db, schema, "users", columnName)
+
+				// The table's column count reflects the drop of the new column
+				TableMustHaveColumnCount(t, db, schema, "users", 1)
+
+				// Insert a row to make sure we get a not possible error because there's no identity column
```
andrew-farries, line 316:
```suggestion
				// Insert a row to make sure we get a `backfill.NotPossible` error when the migration is restarted, because there's no identity column
```

pkg/migrations/op_add_column_test.go
```diff
@@ -150,7 +150,25 @@ func TestAddColumn(t *testing.T) {
...
-			wantStartErr: backfill.NotPossibleError{Table: "users"},
+			afterRollback: func(t *testing.T, db *sql.DB, schema string) {
+				// The new column has been dropped from the underlying table
+				columnName := migrations.TemporaryName("age")
+				ColumnMustNotExist(t, db, schema, "users", columnName)
+
+				// The table's column count reflects the drop of the new column
+				TableMustHaveColumnCount(t, db, schema, "users", 1)
+
+				// Insert a row to make sure we get a not possible error because there's no identity column
```
andrew-farries, line 161:
```suggestion
				// Insert a row to make sure we get a `backfill.NotPossible` error when the migration is restarted, because there's no identity column
```

pkg/migrations/op_add_column_test.go
```diff
@@ -150,7 +150,25 @@ func TestAddColumn(t *testing.T) {
 					},
 				},
 			},
-			wantStartErr: backfill.NotPossibleError{Table: "users"},
+			afterRollback: func(t *testing.T, db *sql.DB, schema string) {
```
andrew-farries, line 153: I think the test should have an `afterStart` part too - even if it's very simple.

pkg/migrations/op_add_column_test.go
```diff
@@ -255,6 +273,58 @@ func TestAddColumn(t *testing.T) {
...
+								Name:     "age",
+								Type:     "integer",
+								Nullable: true,
+							},
+							Up: "1",
+						},
+					},
+				},
+			},
+			afterRollback: func(t *testing.T, db *sql.DB, schema string) {
```
andrew-farries, line 308: I think the test should have an `afterStart` part too - even if it's very simple.

andrew-farries (commented):
The change looks good but I wonder if the ability to skip backfills on empty tables is worth losing the validation check for though?

Optimising for the empty-table case might not be the way to go; when a migration like this is applied to production the table typically won't be empty; and even if it is the same change can be made using a raw SQL migration to avoid the backfill anyway.

We risk allowing writing migrations that will work locally in a dev environment because there is no data in some tables, but then those same migrations might fail against production, without any validation to catch the problem.

Do you have a view here @exekias ?

agedemenli:
@andrew-farries Agree on your concerns. I have tried to add the same empty table check to the validate phase, however, it doesn't look good in terms of readability and structure. 

I'm ok with both merging this PR as is, or closing it.

exekias:
Yeah, it seems risky to count on the number of rows calculation as it is. At the same time, from the bug description, I wonder if we can make assumptions on the table data to state whether a migration is valid. The same migration history could be replayed to a different database, and because of timing, fail this time because the table has rows. In general, we should be able to raise this validation error before touching the database or knowing about its state.

I would be inclined to leave this as a known issue (close the PR) rather than going for a fix that could harm the general use case, WDYT?

andrew-farries:
I think we have consensus that this PR can be closed @agedemenli

PR #649, opened 2025-02-03
Add support for `NOT VALID` in constraint writer

kvch:
This PR adds support for creating not validated constraints in `ALTER TABLE` statements.
I am adding unit tests in a follow-up PR, when `ConstraintSQLWriter` is moved into a separate file.
Either way, it is not used anywhere yet to avoid conflicts with other FK related
changes in future PRs.

Extracted from https://github.com/xataio/pgroll/pull/628

### DIFF
--- a/pkg/migrations/op_create_table.go
+++ b/pkg/migrations/op_create_table.go
@@ -336,6 +336,7 @@ type ConstraintSQLWriter struct {
 	Columns           []string
 	InitiallyDeferred bool
 	Deferrable        bool
+	SkipValidation    bool
 
 	// unique, exclude, primary key constraints support the following options
 	IncludeColumns    []string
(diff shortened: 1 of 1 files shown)


pkg/migrations/op_create_table.go
```diff
@@ -336,6 +336,7 @@ type ConstraintSQLWriter struct {
 	Columns           []string
 	InitiallyDeferred bool
 	Deferrable        bool
+	SkipValidation    bool
```
andrew-farries, line 339: Can this be:
```suggestion
	NotValid    bool
```

so that it more closely maps to the SQL being generated?

kvch: The name of the attribute in PostgreSQL is called `skip_validation` and in go `SkipValidation`: https://github.com/xataio/pg_query_go/blob/4fa82ad6d0369aab38c73903edf45287ddc31cd9/parser/include/postgres/nodes/parsenodes.h#L2684
Hence, the name. Do you still want me to change the name?

andrew-farries: Let's leave it then 😄

andrew-farries (approved):

PR #652, opened 2025-02-04

andrew-farries:
Backfill only rows present at backfill start. This is third approach to solving https://github.com/xataio/pgroll/issues/583. The previous two are:
* https://github.com/xataio/pgroll/pull/634
* https://github.com/xataio/pgroll/pull/648

This is the most direct approach to solving the problem. At the same time as the up/down triggers are created to perform a backfill, a `_pgroll_needs_backfill` column is also created on the table to be backfilled. The column has a `DEFAULT` of `true`; the constant default ensures that this extra column can be added quickly without a lengthy `ACCESS_EXCLUSIVE` lock. The column is removed when the the operation is rolled back or completed.

The up/down triggers are modified to set `_pgroll_needs_backfill` to false whenever they update a row.

The backfill itself is updated to select only rows having `_pgroll_needs_backfill` set to `true` - this ensures that only rows created before the triggers were installed are updated by the backfill. The backfill process still needs to *read* every row in the table, including those inserted/updated after backfill start, but only those rows created before backfill start will be updated.

The main disadvantage of this approach is that backfill now requires an extra column to be created on the target table.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -159,6 +159,8 @@ After some progress updates you should see a message saying that the migration h
   <summary>What's happening behind the progress updates?</summary>
   
   In order to add the new `description` column, `pgroll` creates a temporary `_pgroll_new_description` column and copies over the data from the existing `description` column, using the `up` SQL from the migration. As we have 10^5 rows in our table, this process takes some time. This process is called _backfilling_ and it is performed in batches to avoid locking all rows in the table simultaneously.
+
+  The `_pgroll_needs_backfill` column in the `users` table is used to track which rows have been backfilled and which have not. This column is set to `true` for all rows at the start of the migration and set to `false` once the row has been backfilled. This ensures that rows inserted or updated while the backfill is in process aren't backfilled twice.
 </details>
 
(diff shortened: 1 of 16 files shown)


kvch (commented):
The code itself looks good. The only note I have is rather a modelling question. Could you please explain why you put removing the backfill column to the operations? To me it feels like it should be the responsibility of the backfill process to clean up after itself. So this functionality should be moved to the `backfill` package and should be run from `Roll.Complete` and/or `Roll.Rollback`.

andrew-farries:
Removing the backfill column is in the individual operations because they have the knowledge of which table they are operating on. `Roll.Complete` and `Roll.Rollback` don't have that same information.

We could change this in future and centralize the responsiblity of cleaning up tables (removing functions, triggers and 'needs backfill' columns) into `Roll.Complete` and `Roll.Rollback`, but I think this would be a larger change. It could be worth doing thoug as we have lots of repetition between individual `Complete` and `Rollback`  methods in different operations now.

kvch (approved):
Thank you for fixing this issue and for the detailed explanation.

PR #653, opened 2025-02-04
Expose foreign key constraint options in inline FK definition and in `create_constraint`

kvch:
Now more foreign key options are supported:
* `match_type` in column definitions and in `create_constraint`
* `on_update` in column definitions and in `create_constraint`
* `on_delete_set_columns` in `create_constraint`
* `on_update` in inline FK definitions in a column's `references`
* `match_type` in inline FK definction in a column's `references`

Extracted from https://github.com/xataio/pgroll/pull/628. The only remaining part of that PR is adding support in `sql2pgroll`. :tada:

### DIFF
--- a/pkg/migrations/op_create_table.go
+++ b/pkg/migrations/op_create_table.go
@@ -404,6 +404,7 @@ func (w *ConstraintSQLWriter) WriteForeignKey(referencedTable string, referenced
 	if w.Name != "" {
 		constraint = fmt.Sprintf("CONSTRAINT %s ", pq.QuoteIdentifier(w.Name))
 	}
+	// in case of in line foreign key constraint, columns are already included in the column definition
 	if len(w.Columns) != 0 {
 		constraint += fmt.Sprintf("FOREIGN KEY (%s) ", strings.Join(quoteColumnNames(w.Columns), ", "))
 	}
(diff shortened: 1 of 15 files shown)


kvch:
I am adding more tests.

kvch:
I fixed up the commits and added tests.

pkg/migrations/op_create_table.go
```diff
@@ -404,8 +404,10 @@ func (w *ConstraintSQLWriter) WriteForeignKey(referencedTable string, referenced
 	if w.Name != "" {
 		constraint = fmt.Sprintf("CONSTRAINT %s ", pq.QuoteIdentifier(w.Name))
 	}
-	constraint += fmt.Sprintf("FOREIGN KEY (%s) REFERENCES %s (%s) MATCH %s ON DELETE %s ON UPDATE %s",
-		strings.Join(quoteColumnNames(w.Columns), ", "),
+	if len(w.Columns) != 0 {
```
andrew-farries, line 407: Under what circumstances would we have 0 columns here?

andrew-farries (approved):

pkg/migrations/op_create_table.go
kvch: In case of inline FK constraints. I am adding a comment so it's clear.

PR #658, opened 2025-02-05
Add exclusion constraints to `analyze`/`read_schema`

kvch:
This PR adds support for displaying exclusion constraints in the output of `analyze`/`read_schema`.

Attributes:
* `name`: name of the constraint
* `method`: selected index method (`exclude.index_method` option)
* `columns`: the list of columns included in the constraint
* `predicate`: conditional expression if partials index
* `definition`: constraint definition

```
go run . analyze
{
...
  "excludeConstraints": {
    "forbid_duplicated_titles": {
      "name": "forbid_duplicated_titles",
      "method": "btree",
      "predicate": "(title IS NOT NULL)",
      "columns": [
        "title"
      ],
      "definition": "EXCLUDE USING btree (title WITH =) WHERE ((title IS NOT NULL))"
    }
  },
...
}
```

Furthermore, I added a new attribute called `exclusion` to indexes to show if the index belongs to an exclusion constraint.


pkg/schema/schema.go
```diff
@@ -162,14 +165,17 @@ type ExcludeConstraint struct {
 	// Name is the name of the exclude constraint in postgres
 	Name string `json:"name"`
 
-	// IndexMethod is the index method of the exclude constraint
-	IndexMethod string `json:"indexMethod"`
-
-	// Elements is the elements of the exclude constraint
-	Elements string `json:"elements"`
+	// Method is the index method of the exclude constraint
```
kvch, line 168: Renamed to `Method` so it is consistent with `Method` in `indexes`.

andrew-farries (approved):

PR #660, opened 2025-02-06
Move backfill options to `start` and `migrate` command

### LINKED ISSUE #657: Move `backfill-batch-delay` and `backfill-batch-size` flags to the `start` command
Currently, the `--backfill-batch-delay` and `--backfill-batch-size` flags are global flags:

```
Usage:
  pgroll [command]

Available Commands:
  complete    Complete an ongoing migration with the operations present in the given file
  completion  Generate the autocompletion script for the specified shell
  help        Help about any command
  init        Initialize pgroll in the target database
  latest      Print the name of the latest schema version, either in the target database or a local directory
  migrate     Apply outstanding migrations from a directory to a database
  pull        Pull migration history from the target database and write it to disk
  rollback    Roll back an ongoing migration
  start       Start a migration for the operations present in the given file
  status      Show pgroll status

Flags:
      --backfill-batch-delay duration   Duration of delay between batch backfills (eg. 1s, 1000ms)
      --backfill-batch-size int         Number of rows backfilled in each batch (default 1000)
  -h, --help                            help for pgroll
      --lock-timeout int                Postgres lock timeout in milliseconds for pgroll DDL operations (default 500)
      --pgroll-schema string            Postgres schema to use for pgroll internal state (default "pgroll")
      --postgres-url string             Postgres URL (default "postgres://postgres:postgres@localhost?sslmode=disable")
      --role string                     Optional postgres role to set when executing migrations
      --schema string                   Postgres schema to use for the migration (default "public")
  -v, --version                         version for pgroll

Use "pgroll [command] --help" for more information about a command.
```

but they only make sense when applied to the `start` command.

The two flags should be removed from global flags and added as `start` command flags.

kvch:
This PR moves the flags `backfill-batch-delay` and `backfill-batch-size` to
the commands `start` and `migrate`. So these options no longer show up in the list
of global flags.

I created a new struct `backfill.Config` to store and pass the options down
to the migration runner. I also moved the default values for batch delay and batch size
into the package `backfill`. I renamed `options.go` to `config.go` because it is a better name given the new contents of the file.

Closes #657


cli-definition.json
```diff
@@ -141,16 +141,6 @@
     }
   ],
   "flags": [
-    {
```
andrew-farries, line 144: I'd have expected these flags to have moved, not just removed.

I wonder if the tool that generates this CLI description is working correctly.

pkg/backfill/config.go
```diff
@@ -0,0 +1,66 @@
...
+func WithCallbacks(cbs ...CallbackFn) OptionFn {
+	return func(o *Config) {
+		o.callbacks = cbs
+	}
+}
+
+func (c *Config) AddCallback(fn CallbackFn) {
+	if c.callbacks == nil {
+		c.callbacks = make([]CallbackFn, 0)
+	}
```
andrew-farries, line 60: Not needed I think - appending to a `nil` slice works fine.

pkg/backfill/config.go
```diff
@@ -0,0 +1,66 @@
...
+// WithBatchDelay sets the delay between batches for the backfill operation.
+func WithBatchDelay(delay time.Duration) OptionFn {
+	return func(o *Config) {
+		o.batchDelay = delay
+	}
+}
+
+// WithCallbacks sets the callbacks for the backfill operation.
+// Callbacks are invoked after each batch is processed.
+func WithCallbacks(cbs ...CallbackFn) OptionFn {
```
andrew-farries, line 51: Could remove this and change its only usage to `AddCallback` instead. Or remove `AddCallback` and update its usage to `WithCallbacks`. We don't need both methods IMO.

cmd/migrate.go
```diff
@@ -59,20 +63,30 @@ func migrateCmd() *cobra.Command {
...
 			}
 
 			// Run the final migration, completing it only if requested.
-			return runMigration(ctx, m, migs[len(migs)-1], complete)
+			return runMigration(ctx, m, migs[len(migs)-1], complete, backfillConfig)
 		},
 	}
 
+	migrateCmd.PersistentFlags().Int("backfill-batch-size", backfill.DefaultBatchSize, "Number of rows backfilled in each batch")
+	migrateCmd.PersistentFlags().Duration("backfill-batch-delay", backfill.DefaultDelay, "Duration of delay between batch backfills (eg. 1s, 1000ms)")
```
andrew-farries, line 85: These don't need to be persistent flags (they don't apply to subcommands, `migrate` doesn't have any). use `.Flags()` instead, like the `--complete` flag.

This might explain why the `cli-description.json` didn't notice these flags.

cmd/start.go
```diff
@@ -34,30 +35,39 @@ func startCmd() *cobra.Command {
...
+				backfill.WithBatchSize(flags.BackfillBatchSize()),
+				backfill.WithBatchDelay(flags.BackfillBatchDelay()),
+			)
+
+			return runMigrationFromFile(cmd.Context(), m, fileName, complete, c)
 		},
 	}
 
+	startCmd.PersistentFlags().Int("backfill-batch-size", backfill.DefaultBatchSize, "Number of rows backfilled in each batch")
+	startCmd.PersistentFlags().Duration("backfill-batch-delay", backfill.DefaultDelay, "Duration of delay between batch backfills (eg. 1s, 1000ms)")
```
andrew-farries, line 48: Same comment about not needing to use `PersistentFlags` here as for the `migrate` command.

pkg/backfill/backfill.go
```diff
@@ -15,24 +15,18 @@ import (
 )
 
 type Backfill struct {
-	conn       db.DB
-	batchSize  int
-	batchDelay time.Duration
-	callbacks  []CallbackFn
+	conn   db.DB
+	config *Config
```
andrew-farries, line 19: The `config` struct could be embedded: 

```suggestion
   *Config
```

I don't mind either way.

pkg/roll/execute_test.go
```diff
@@ -33,7 +34,7 @@ func TestSchemaIsCreatedAfterMigrationStart(t *testing.T) {
 		ctx := context.Background()
 		version := "1_create_table"
 
-		if err := mig.Start(ctx, &migrations.Migration{Name: version, Operations: migrations.Operations{createTableOp("table1")}}); err != nil {
+		if err := mig.Start(ctx, &migrations.Migration{Name: version, Operations: migrations.Operations{createTableOp("table1")}}, backfill.NewConfig()); err != nil {
```
andrew-farries, line 37: The signature of `Start` could change to:

```go
func (m *Roll) Start(ctx context.Context, migration *migrations.Migration, c ...*backfill.Config) error
```

so that the config becomes optional and we don't have to do this everywhere.

It's arguably an anti-pattern though. WDYT?

andrew-farries (commented):
I think this has broken the flags - they aren't taking effect for me:

Apply the migration:
```json
{
  "name": "01_create_table",
  "operations": [
    {
      "create_table": {
        "name": "items",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "name",
            "type": "text"
          }
        ]
      }
    }
  ]
}
```

Fill the table with data:

```sql
insert into items(name) select 'name_' || i from generate_series(1, 1_000_000) as i
```

Start a migration with a specified delay:

```bash
go run . start --backfill-batch-delay=10s 02_set_unique.json
```

where the migration is defined:

```json
{
  "name": "02_set_unique",
  "operations": [
    {
      "alter_column": {
        "table": "items",
        "column": "name",
        "unique": {
          "name": "items_name_unique"
        },
        "up": "name",
        "down": "name"
      }
    }
  ]
}
```

There is no delay between each batch during the backfill.

I think the problem is how the flags are defined as `PersistentFlags` in the `start` and `migrate` commands.

pkg/backfill/config.go
kvch: Copilot :anger:

andrew-farries (commented):
The flags still don't work for me, using the same reproduction as in this comment:

https://github.com/xataio/pgroll/pull/660#pullrequestreview-2601443317

pkg/backfill/config.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// WithBatchDelay sets the delay between batches for the backfill operation.
+func WithBatchDelay(delay time.Duration) OptionFn {
+	return func(o *Config) {
+		o.batchDelay = delay
+	}
+}
+
+func (c *Config) AddCallback(fn CallbackFn) {
```
andrew-farries, line 49: Other options have doc comments, so this one could have one too

pkg/roll/execute.go
```diff
@@ -16,14 +16,14 @@ import (
 )
 
 // Start will apply the required changes to enable supporting the new schema version
-func (m *Roll) Start(ctx context.Context, migration *migrations.Migration, cbs ...backfill.CallbackFn) error {
+func (m *Roll) Start(ctx context.Context, migration *migrations.Migration, c *backfill.Config) error {
```
andrew-farries, line 19: Calling the parameter `cfg` would be better IMO.

pkg/roll/execute.go
```diff
@@ -309,11 +309,8 @@ func (m *Roll) ensureView(ctx context.Context, version, name string, table *sche
 	return nil
 }
 
-func (m *Roll) performBackfills(ctx context.Context, tables []*schema.Table, cbs ...backfill.CallbackFn) error {
-	bf := backfill.New(m.pgConn,
-		backfill.WithBatchSize(m.backfillBatchSize),
-		backfill.WithBatchDelay(m.backfillBatchDelay),
-		backfill.WithCallbacks(cbs...))
+func (m *Roll) performBackfills(ctx context.Context, tables []*schema.Table, c *backfill.Config) error {
```
andrew-farries, line 312: Same comment about changing parameter name to `cfg`

andrew-farries (approved):
Looks good now; the flags work as expected 👍 

Left some comments, feel free to address them or not.

PR #665, opened 2025-02-07
Move `ConstraintSQLWriter` to a separate file

kvch:
This PR moves `ConstraintSQLWriter` from `op_create_table` into its own file.
The struct is used in multiple operations, so it belongs to a dedicated file.

I added comments and unit tests.


andrew-farries (approved):
👍 Thanks for adding all the comments 👍

PR #666, opened 2025-02-07
Improve validation of `create_table` and `add_column` operations

andrew-farries:
Improve validation of `create_table` and `add_column` operations.

Previously, these operations would fail at migration `Start` if one of the column definitions was invalid due to having a missing `name` or `type` field.

This PR adds extra validation checks to both operations to ensure that such migrations fail at the validation stage instead.

### DIFF
--- a/pkg/migrations/column.go
+++ b/pkg/migrations/column.go
@@ -26,3 +26,15 @@ func (c *Column) HasImplicitDefault() bool {
 		return false
 	}
 }
+
+// Validate returns true iff the column contains all fields required to create
+// the column
+func (c *Column) Validate() bool {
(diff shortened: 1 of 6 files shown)


andrew-farries:
> These errors are caught by the JSON validator when the user writes their config files, no?

Yes, the JSON schema will reject these errors. 

We should use JSON schema validation in `pgroll` itself - currently it's not used. It was added to support third party tools like the migration editor in Xata, but `pgroll` itself doesn't use the JSON schema to validate migration files.

I created https://github.com/xataio/pgroll/issues/673 to start doing JSON validation in `pgroll` itself.

Issue #664, opened 2025-02-07
migrate command fails when there's an inferred migration


agedemenli:
To reproduce, after `pgroll init`, add some inferred migration, e.g simply by running `CREATE TABLE t1 (a int)`. Then any `migrate` command will fail with the error:
```
Error: failed to get migrations to apply: remote migration does not match local migration: remote="sql_5b73a72b63fb73", local="01_create_tables"
```
The inferred sql migration is obviously not in the list of migrations to be applied, so we get the mismatch error.

We should maybe consider excluding `inferred` migrations from schema history.

PR #667, opened 2025-02-07
Ignore inferred migrations for `migrate` and `pull` commands

agedemenli:
`SchemaHistory` function returns list of all migrations applied on a schema, including inferred ones. This causes `migrate` command to fail with a migration mismatch error, as described in #664. This PR ignores inferred migrations so that case will be fixed. Also, inferred migrations will not be included in the output of `pull` command as well.

Though, I'm not sure if we should exclude inferred migrations from `pull`. We can also keep them for `pull` while excluding only from `migrate`.

### DIFF
--- a/pkg/roll/unapplied.go
+++ b/pkg/roll/unapplied.go
@@ -17,11 +17,6 @@ import (
 // If the local order of migrations does not match the order of migrations in
 // the schema history, an `ErrMismatchedMigration` error is returned.
 func (m *Roll) UnappliedMigrations(ctx context.Context, dir fs.FS) ([]*migrations.Migration, error) {
-	latestVersion, err := m.State().LatestVersion(ctx, m.Schema())
-	if err != nil {
-		return nil, fmt.Errorf("determining latest version: %w", err)
-	}
-
 	files, err := fs.Glob(dir, "*.json")
 	if err != nil {
 		return nil, fmt.Errorf("reading directory: %w", err)
@@ -34,7 +29,7 @@ func (m *Roll) UnappliedMigrations(ctx context.Context, dir fs.FS) ([]*migration
 
 	// Find the index of the first unapplied migration
 	var idx int
-	if latestVersion != nil {
+	if len(history) > 0 {
 		for _, file := range files {
 			migration, err := openAndReadMigrationFile(dir, file)
 			if err != nil {
@@ -47,7 +42,7 @@ func (m *Roll) UnappliedMigrations(ctx context.Context, dir fs.FS) ([]*migration
 			}
 
 			idx++
-			if migration.Name == *latestVersion {
+			if idx == len(history) {
 				break
 			}
 		}
--- a/pkg/state/history.go
+++ b/pkg/state/history.go
@@ -28,7 +28,7 @@ func (s *State) SchemaHistory(ctx context.Context, schema string) ([]Migration, 
 	rows, err := s.pgConn.QueryContext(ctx,
 		fmt.Sprintf(`SELECT name, migration, created_at
 			FROM %s.migrations
-			WHERE schema=$1 ORDER BY created_at`,
+			WHERE schema=$1 AND migration_type='pgroll' ORDER BY created_at`,
 			pq.QuoteIdentifier(s.schema)), schema)
 	if err != nil {
 		return nil, err


andrew-farries:
I think the desired workflow here might be:

1. User runs `pgroll migrate` 
2. User hits error that local schema history doesn't match remote history.
3. User runs `pgroll pull` to pull down the complete schema history 
4. User resolves whatever issues arise, commits new migrations to source control.
5. User runs `pgroll migrate`.

I think the `inferred` migration case is no different to the case where it's a `pgroll` migration causing the history mismatch - if we want the complete history of the schema, then `inferred` migrations need to be included locally too.

WDYT @exekias?  I know we talked previously about whether we want to persist with capturing DDL as inferred migrations, but as long as we do I think we don't want to ignore them when running `migrate`.

exekias:
Yeah, that would make sense to me, I'm also curious about @divyenduz thoughts here

divyenduz:
Are inferred migration the one that `pgroll` tracks by tracking all DDL that happen directly on a connection string? IMO what Andrew described should be the approach. As long as we infer a migration, pull / migrate should use/return it like any other migration. 

This would mean, manual resolution from the user, in case local / remote histories don't align.

agedemenli:
> commits new migrations to source control

How does user commit a new migration to make sure local migrations are the same with remote history? Like adding a `pgroll` migration but naming it like an inferred one like `sql_<14digit hexadecimal>`?

andrew-farries:
Here's how the flow would work, using the `examples/` migrations from this repo and an empty DB with `pgroll` initialized:

* Remove the final migration file from the `examples/` dir:

```bash
rm examples/52_create_table_with_exclusion_constraint.json
```

* Apply migrations from the `examples/` dir:

```bash
go run . migrate --complete examples/ 
```

* Create an inferred migration by adding a table directly:

```sql
CREATE TABLE foo(a int)
```

* Check the `pgroll.migrations` table looks as expected:

```sql
select * from pgroll.migrations where schema='public' order by created_at desc
```

```
+--------+---------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
| schema | name                                              | migration                                                                                                                                                                                                                                    >
|--------+---------------------------------------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------->
| public | sql_18e81034a8cef6                                | {"name": "sql_18e81034a8cef6", "operations": [{"sql": {"up": "CREATE TABLE foo(a int)"}}]}                                                                                                                                                   >
| public | 51_create_table_with_table_foreign_key_constraint | {"name": "51_create_table_with_table_foreign_key_constraint", "operations": [{"create_table": {"name": "phonebook", "columns": [{"name": "id", "type": "serial"}, {"name": "provider_id", "type": "serial"}, {"name": "name", "type": "varcha>
| public | 50_create_table_with_table_constraint             | {"name": "50_create_table_with_table_constraint", "operations": [{"create_table": {"name": "telephone_providers", "columns": [{"name": "id", "type": "serial"}, {"name": "name", "type": "varchar(255)"}, {"name": "tax_id", "type": "varchar>
| public | 49_unset_not_null_on_indexed_column               | {"name": "49_unset_not_null_on_indexed_column", "operations": [{"alter_column": {"up": "SELECT CASE WHEN name IS NULL THEN 'unknown fruit' ELSE name END", "down": "name", "table": "fruits", "column": "name", "nullable": true}}]}   
```

* Restore the missing migration file from the `examples/` dir

```bash
git reset --hard
```

* Attempt to apply the final migration (`examples/52_create_table_with_exclusion_constraint.json`):

```bash
go run . migrate --complete examples/
```

This fails with an error:

```
Error: failed to get migrations to apply: remote migration does not match local migration: remote="sql_18e81034a8cef6", local="52_create_table_with_table_exclusion_constraint"
```

* Pull the remote migrations into a temporary directory:

```
go run . pull -p tmp
```

At this point, the user has to decide how to proceed; there is an extra migration on the remote (`tmp/0052_sql_18e81034a8cef6.json`) that isn't in the local `examples/` dir. 

* We decide to insert the new (`0052_sql_18e81034a8cef6.json`) migration before our `examples/52_create_table_with_exclusion_constraint.json` migration:

```
mv tmp/0052_sql_18e81034a8cef6.json examples/52_create_foo_table.json 
mv examples/52_create_table_with_exclusion_constraint.json examples/53_create_table_with_exclusion_constraint.json
```

* Now the local and remote migration histories match again and `migrate` works:

```
go run . migrate --complete examples/
 SUCCESS  New version of the schema available under the postgres "public_52_create_table_with_table_exclusion_constraint" schema
```

---

This is similar to a `git` rebase workflow where the upstream (the set of migrations applied to the remote) has diverged from our local branch (the set of migrations we have on disk).

PR #668, opened 2025-02-10
Alter sequence owner before dropping the column

agedemenli:


kvch (changes requested):
Unfortunately, the bug can surface in all migration types if the column that has a serial sequence is being duplicated. Thus, it is not enough to change the owner only in create constraint operation. The issue must be addressed in `backfill.go` and/or `rename.go` and in more migration types.

kvch (changes requested):
Looks good. Do you mind adding a few more tests for dropping multi-column constraints and altering columns?

kvch (approved):

PR #670, opened 2025-02-10
Add support for `no_inherit` on column level constraints and in `create_constraint` operation

kvch:
This PR adds a new option `no_inherit` to inline and table level check constraints in `create_constraint`. It also makes sure that the check is duplicated correctly during migration. Also, I added the new attribute `noInherit` to the output of `analyze`/`read_schema`.

### DIFF
--- a/docs/operations/create_constraint.mdx
+++ b/docs/operations/create_constraint.mdx
@@ -17,6 +17,7 @@ Required fields: `name`, `table`, `type`, `up`, `down`.
     "columns": ["column1", "column2"],
     "type": "unique"| "check" | "foreign_key",
     "check": "SQL expression for CHECK constraint",
+    "no_inherit": "true|false",
     "references": {
       "name": "name of foreign key reference",
       "table": "name of referenced table",
--- a/internal/jsonschema/testdata/create-constraint-1-invalid-check.txtar
+++ b/internal/jsonschema/testdata/create-constraint-1-invalid-check.txtar
@@ -1,0 +1,28 @@
+This is an invalid 'create_constraint' migration.
+Check constraint must have the option 'check' configured.
+
+-- create_constraint.json --
+{
+  "name": "migration_name",
+  "operations": [
+    {
+      "create_constraint": {
+        "name": "my_invalid_check",
+        "table": "my_table",
+        "type": "check",
+        "columns": [
+          "my_column"
+        ],
+        "up": {
+          "my_column": "my_column"
+        },
+        "down": {
+          "my_column": "my_column"
+        }
+      }
+    }
+  ]
+}
+
+-- valid --
+false
--- a/pkg/migrations/duplicate_test.go
+++ b/pkg/migrations/duplicate_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	pgq "github.com/xataio/pg_query_go/v6"
 
 	"github.com/xataio/pgroll/pkg/schema"
 )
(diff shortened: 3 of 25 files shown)


pkg/migrations/op_create_table.go
```diff
@@ -363,7 +364,11 @@ func (w *ConstraintSQLWriter) WriteCheck(check string, noInherit bool) string {
 	if w.Name != "" {
 		constraint = fmt.Sprintf("CONSTRAINT %s ", pq.QuoteIdentifier(w.Name))
 	}
-	constraint += fmt.Sprintf("CHECK (%s)", check)
+	if !strings.HasPrefix(check, "CHECK (") {
```
kvch, line 367: I added this step so we can use the constraint writer in the duplication process as well.

docs/operations/create_constraint.mdx
```diff
@@ -17,6 +17,7 @@ Required fields: `name`, `table`, `type`, `up`, `down`.
     "columns": ["column1", "column2"],
     "type": "unique"| "check" | "foreign_key",
     "check": "SQL expression for CHECK constraint",
+    "no_inherit": "true|false",
```
andrew-farries, line 20: Should this also mention that it's only valid for `CHECK` constraints?

internal/jsonschema/testdata/create-constraint-1-invalid-check.txtar
```diff
@@ -0,0 +1,28 @@
+This is an invalid 'create_constraint' migration.
+Check constraint must have the option 'check' configured.
```
andrew-farries, line 2: Do we need further schema tests to check that `no_inherit` is valid only for `CHECK` constraints?

pkg/migrations/duplicate_test.go
```diff
@@ -67,35 +68,37 @@ func TestDuplicateStmtBuilderCheckConstraints(t *testing.T) {
...
-			expectedStmts: []string{`ALTER TABLE "test_table" ADD CONSTRAINT "_pgroll_dup_different_nick" "_pgroll_new_name" != "_pgroll_new_nick" NOT VALID`},
+			expectedStmts: []string{`ALTER TABLE "test_table" ADD CONSTRAINT "_pgroll_dup_different_nick" CHECK ("_pgroll_new_name" != "_pgroll_new_nick") NOT VALID`},
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
 			stmts := d.duplicateCheckConstraints(nil, testCases.columns...)
 			assert.Equal(t, len(testCases.expectedStmts), len(stmts))
 			for _, stmt := range stmts {
+				_, err := pgq.Parse(stmt)
+				assert.NoError(t, err)
```
andrew-farries, line 101: What's the reason for this change?

pkg/migrations/op_create_table.go
andrew-farries: The `ConstraintWriter` has been moved into a separate file by an already-merged PR right (#665)?  

Maybe this PR needs to be rebased onto `main`?

andrew-farries: Do the tests that were added in #665 need to be updated to test this new check?

andrew-farries (commented):
Changes look OK, but I think the PR needs to be rebased to pick up relevant changes from `main`, eg #665.

Table inheritance feels like a very rarely used Postgres feature; did this work arise from seeing it used somwhere like an ORM migration, or for completeness to cover all constraint options? In general, I think we need to be guided more by the former and add features only when we see them in use.

pkg/migrations/duplicate_test.go
kvch: The statements that were generated previously had SQL syntax errors. It was never a problem for pgroll, it never caused a bug. But I expected that this statement generator produces a valid SQL statement.

kvch:
I won't close this PR, it comes with some nice cleanups.

docs/operations/create_constraint.mdx
kvch: We don't mention it in the description of `constraint`, so I omitted it here as well.

andrew-farries (approved):

Issue #644, opened 2025-01-31
Adding columns with `UNIQUE` takes an exclusive table lock


andrew-farries:
Adding a new column with `unique: true` takes a long-lived `ACCESS_EXCLUSIVE` lock on the target table.

The `unique` condition on the new column should be added in the same way that uniqueness is added to existing columns:
* On start, create a `UNIQUE` index on the new column concurrently.
* On complete, create a `UNIQUE` constraint using the index.

This two-step approach avoids the lock.

## Reproduction

1. Create a table `items`:

```json
{
  "name": "01_create_table",
  "operations": [
    {
      "create_table": {
        "name": "items",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "name",
            "type": "text"
          }
        ]
      }
    }
  ]
}
```

2. Populate it with some 10s of millions of rows

```sql
INSERT INTO items(name) SELECT 'name_' || i FROM generate_series(1, 200000) AS i
```

3. Add a column with a `unique: true`:

```json
{
  "name": "02_add_column",
  "operations": [
    {
      "add_column": {
        "table": "items",
        "up": "random()::text",
        "column": {
          "name": "foo",
          "type": "text",
          "unique": true
        }
      }
    }
  ]
}
```

The `items` table is now locked at migration start:

```sql
SELECT * from pg_locks WHERE relation = 'items'::regclass;
```

PR #679, opened 2025-02-12
Add unique column without exclusive lock

mcadariu:
Hi, 

Awesome project! 👍 
This PR is an attempt at fixing https://github.com/xataio/pgroll/issues/644.

Looking forward to your feedback!


pkg/migrations/op_add_column.go
```diff
@@ -42,6 +42,12 @@ func (o *OpAddColumn) Start(ctx context.Context, conn db.DB, latestSchema string
 		}
 	}
 
+	if o.Column.Unique {
```
kvch, line 45: In `addColumn` we already add `UNIQUE` to the generated SQL. You also need to change that function, so `pgroll` does not add unique constraint directly to the column. Similarly to what we do for check constraints.

kvch (commented):
Great start! I added a single note. Thank you for your contribution! :tada:

agedemenli:
Thanks for the contribution!

Just tested it with
```
./pgroll init
./pgroll start ./examples/01_create_tables.json --complete
./pgroll start ./examples/02_create_another_table.json --complete 
./pgroll start ./examples/03_add_column.json --complete 
```
after adding `"unique": true` for all three columns added in `03_add_column.json`
Ended up with below output:
```
\d+ products
                                                                 Table "public.products"
   Column    │          Type          │ Collation │ Nullable │               Default                │ Storage  │ Compression │ Stats target │ Description
─────────────┼────────────────────────┼───────────┼──────────┼──────────────────────────────────────┼──────────┼─────────────┼──────────────┼─────────────
 id          │ integer                │           │ not null │ nextval('products_id_seq'::regclass) │ plain    │             │              │
 name        │ character varying(255) │           │ not null │                                      │ extended │             │              │
 price       │ numeric(10,2)          │           │ not null │                                      │ main     │             │              │
 description │ character varying(255) │           │          │                                      │ extended │             │              │
 stock       │ integer                │           │ not null │ 100                                  │ plain    │             │              │
 category    │ character varying(255) │           │ not null │                                      │ extended │             │              │
Indexes:
    "products_pkey" PRIMARY KEY, btree (id)
    "products__pgroll_new_category_key" UNIQUE CONSTRAINT, btree (category)
    "products__pgroll_new_description_key" UNIQUE CONSTRAINT, btree (description)
    "products__pgroll_new_stock_key" UNIQUE CONSTRAINT, btree (stock)
    "products_category_key" UNIQUE CONSTRAINT, btree (category)
    "products_description_key" UNIQUE CONSTRAINT, btree (description)
    "products_name_key" UNIQUE CONSTRAINT, btree (name)
    "products_stock_key" UNIQUE CONSTRAINT, btree (stock)
Access method: heap
```

Looks like now we have double unique constraint for each unique column, one with the actual column name, and one with the temporary one.

pkg/migrations/op_add_column.go
```diff
@@ -268,6 +285,11 @@ func (o *OpAddColumn) addCheckConstraint(ctx context.Context, tableName string,
 	return err
 }
 
+// UniqueIndexName returns the name of the unique index for the given column
+func UniqueIndexName(columnName string) string {
+	return columnName + "uniq"
+}
```
andrew-farries, line 291: Elsewhere, these temporary names use `_pgroll` prefixes. We should follow that patten here too:

```suggestion
func UniqueIndexName(columnName string) string {
	return "_pgroll_uniq_"  + columnName + "uniq"
}
```

pkg/migrations/op_add_column_test.go
```diff
@@ -111,6 +111,82 @@ func TestAddColumn(t *testing.T) {
 				}, res)
 			},
 		},
+		{
+			name: "add column unique",
```
andrew-farries, line 115: I don't think this test adds any new coverage; it passes before and after the changes in the rest of this PR.

These tests and the ones like it in the `op_*_test.go` files are tests for the correctness of the various `pgroll` operations and work with small amounts of data. Issues like https://github.com/xataio/pgroll/issues/644 are only visible when we're working with much larger row counts.

Lock backoff and retry is tested elsewhere (like in `execute_test.go`), but there are no automated tests for this kind of two step constraint validation for other things like ensuring adding `NOT NULL` constraint doesn't lock the table for an extended period of time.

This highlights a gap in our testing, but I don't think we should address it in this PR.

andrew-farries (commented):
Thanks for doing this @mcadariu 👍 

Left some comments, I think it's pretty close.

pkg/migrations/op_add_column_test.go
mcadariu: Fair point! Removed.

mcadariu:
Thanks all! 🙏 

Ran Ahmet's scenario to try it out. In `./examples/03_add_column.json` I've set all columns to unique: 

```json
{
  "name": "03_add_column_to_products",
  "operations": [
    {
      "add_column": {
        "table": "products",
        "up": "UPPER(name)",
        "column": {
          "name": "description",
          "type": "varchar(255)",
          "nullable": true,
          "unique": true
        }
      }
    },
    {
      "add_column": {
        "table": "products",
        "column": {
          "name": "stock",
          "type": "int",
          "nullable": false,
          "default": "100",
          "unique": true
        }
      }
    },
    {
      "add_column": {
        "table": "products",
        "up": "name || '-category'",
        "column": {
          "name": "category",
          "type": "varchar(255)",
          "nullable": false,
          "unique": true
        }
      }
    }
  ]
}
```

The result looks OK to me: 

```
postgres=# \d products;
                                      Table "public.products"
   Column    |          Type          | Collation | Nullable |               Default                
-------------+------------------------+-----------+----------+--------------------------------------
 id          | integer                |           | not null | nextval('products_id_seq'::regclass)
 name        | character varying(255) |           | not null | 
 price       | numeric(10,2)          |           | not null | 
 description | character varying(255) |           |          | 
 stock       | integer                |           | not null | 100
 category    | character varying(255) |           | not null | 
Indexes:
    "products_pkey" PRIMARY KEY, btree (id)
    "products_category_key" UNIQUE CONSTRAINT, btree (category)
    "products_description_key" UNIQUE CONSTRAINT, btree (description)
    "products_name_key" UNIQUE CONSTRAINT, btree (name)
    "products_stock_key" UNIQUE CONSTRAINT, btree (stock)
```

The `name` column was already set to unique in `02_create_another_table.json`.

kvch (approved):
Thank you the contribution and for making pgroll better! :smile:

gulcin:
Hello @mcadariu, thank you for your contribution to pgroll 🎉 I got so happy to see a familiar face in our project! Have a nice week ahead.

mcadariu:
Hi @gulcin! 
My pleasure! Yes, it's a small world 😄 
Likewise, have a good one.

PR #685, opened 2025-02-14
Add new command `convert` to convert SQL migrations to pgroll operations

kvch:
This PR adds a skeleton command for converting SQL migration files
to pgroll migration files.

The new command looks like this:

```sh
pgroll convert {path-to-sql-file}
```

All the command does is prints the file name and panics because it is
not implemented.


andrew-farries (commented):
We also have the `pgroll sql` hidden command which takes a SQL string as a parameter and converts its. Do you have an opinion on whether we need one or both of these?

Idea:

`pgroll convert` could take either a filename, or if none is provided, read from `stdin`. This would allow piping SQL through `convert` from some other source.

kvch:
> Do you have an opinion on whether we need one or both of these?

I would remove the command `sql` because it seems unnecessary.

> `pgroll convert` could take either a filename, or if none is provided, read from stdin. This would allow piping SQL through `convert` from some other source.

Yes, I also considered this. So I like this idea. I was also thinking if the user configures a folder as an input, `convert` would open all sql files in that folder and tried to translate the migrations into pgroll ops. If the file is not a directory, pgroll opens the file, and translates its contents. I just worry that it might be too confusing with all of these different modes. :D

andrew-farries (approved):

Issue #643, opened 2025-01-31
Adding columns with volatile `DEFAULT`s locks tables for duration of table rewrite


andrew-farries:
Adding a column with a volatile `DEFAULT`  to a table takes an `ACCESS_EXCLUSIVE` lock on the table for the duration of the full table rewrite required to populate the default value into every record.

As of Postgres 11, adding a column with a non-volatile `DEFAULT`[does not require](https://www.enterprisedb.com/blog/adding-new-table-columns-default-values-postgresql-11-0) a table rewrite, but this optimization does not apply to volatile defaults.

This means that the goal of lock-free migrations is not met for this kind of `add_column` operation.

A possible solution is for `add_column` operations with `DEFAULT`s to add the default value manually via an `up` expression and backfill and then add the `DEFAULT` afterwards. This avoids taking the long-lived exclusive lock. This would require some way to distinguish between non-volatile defaults (which are fast to add, and would be slowed down by requiring a backfill) and volatile ones.

## Reproduction

1. Create a table `items`:

```json
{
  "name": "01_create_table",
  "operations": [
    {
      "create_table": {
        "name": "items",
        "columns": [
          {
            "name": "id",
            "type": "serial",
            "pk": true
          },
          {
            "name": "name",
            "type": "text"
          }
        ]
      }
    }
  ]
}
```

2. Populate it with some 10s of millions of rows

```sql
INSERT INTO items(name) SELECT 'name_' || i FROM generate_series(1, 200000) AS i
```

3. Add a column with a volatile default:

```json
{
  "name": "02_add_column",
  "operations": [
    {
      "add_column": {
        "table": "items",
        "column": {
          "name": "foo",
          "type": "int",
          "default": "floor(random() * (50 - 10 + 1) + 10)",
          "nullable": true
        }
      }
    }
  ]
}
```

The `items` table is now locked for as long as it takes for a table rewrite to add the `DEFAULT` value to every row in the table:

```sql
SELECT * from pg_locks WHERE relation = 'items'::regclass;
```

PR #694, opened 2025-02-19
Add columns with volatile `DEFAULT`s without an `ACCESS_EXCLUSIVE` table lock

andrew-farries:
Ensure that adding a new column (with an `add_column` operation) does not take an `ACCESS_EXCLUSIVE` lock on the table even when the column has a volatile `DEFAULT` expression.

Adding a column with a `DEFAULT` expression in Postgres can have one of two different locking behaviours depending on the expression used as the `DEFAULT`:

* For non-volatile `DEFAULT` expressions, the column addition can use the fast-path optimization to add the column default; this is a catalog-only operation and so holds a lock only for a very short time.
* For volatile `DEFAULT` expressions, the fast-path optimization cannot be used; the change requires a full table rewrite - this means holding an `ACCESS_EXCLUSIVE` lock on the table until the rewrite is complete.

This PR makes `add_column` operations aware of this distinction and able to add columns with volatile defaults without taking an `ACCESS_EXCLUSIVE` lock on the table. 

When adding a column with a `DEFAULT`:

* create a temporary schema-only copy of the table.
* add the column to this temporary table
* read the [pg_attribute.atthasmissing](https://www.postgresql.org/docs/current/catalog-pg-attribute.html) system catalog field to determine whether the fast-path optimization was used to add the column `DEFAULT`.
* drop the temporary table

Now the operation knows whether the column can be added using the fast-path `DEFAULT` optimization. In the case where the column can use the optimization, the column can be added with a `DEFAULT` directly. In the case where the fast-path optimization can't be used:

* Add the column without a `DEFAULT` value
* Create an `up` trigger to backfill the column using the column `DEFAULT` expression
* Add the `DEFAULT` to the column on migration completion

This avoids the `ACCESS_EXCLUSIVE` lock; effectively replacing the table rewrite that would have caused it with a backfill operation instead.

If the column to be added has a volatile `DEFAULT` and a  `NOT NULL` constraint, the column must be added with a `NOT NULL ... NOT VALID` constraint to allow the column to be added without a `DEFAULT`; the constraint is validated and upgraded on migration completion.

For a non-volatile `DEFAULT`s, the operation must set `up` to the same expression as is set in `default`. The operation will fail and roll back if they differ.

## References

Volatile vs non-volatile functions:
* https://www.postgresql.org/docs/current/xfunc-volatility.html

The fast-path optimization for non-volatile column `DEFAULT` expressions is explained here:

* https://brandur.org/postgres-default
* https://www.enterprisedb.com/blog/adding-new-table-columns-default-values-postgresql-11

Fixes https://github.com/xataio/pgroll/issues/643

### DIFF
--- a/internal/defaults/fastpath.go
+++ b/internal/defaults/fastpath.go
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package defaults
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/db"
+)
+
+const (
+	cNewTableName  = "_pgroll_temp_fastpath_check"
+	cNewColumnName = "_pgroll_fastpath_check_column"
+)
+
+// UsesFastPath returns true if [defaultExpr] will use the fast-path
+// optimization added in Postgres 11 to avoid taking an `ACCESS_EXCLUSIVE` lock
+// when adding a new column with a `DEFAULT` value.
+//
+// The implementation works by creating a schema-only copy of [tableName],
+// adding a new column with a `DEFAULT` value of [defaultExpr] and checking
+// system catalogs to see if the fast-path optimization was applied.
+func UsesFastPath(ctx context.Context, conn db.DB, tableName, columnType, defaultExpr string) (bool, error) {
+	// Check if we have a real connection or a fake one
+	if _, ok := conn.(*db.FakeDB); ok {
+		return true, nil
+	}
+
+	// Create a schema-only copy of the table
+	_, err := conn.ExecContext(ctx, fmt.Sprintf("CREATE UNLOGGED TABLE %s AS SELECT * FROM %s WHERE false",
+		pq.QuoteIdentifier(cNewTableName),
+		pq.QuoteIdentifier(tableName)))
+	if err != nil {
+		return false, fmt.Errorf("failed to create schema-only copy of table: %w", err)
+	}
+
+	// Ensure that the schema-only copy is removed
+	defer cleanup(ctx, conn)
+
+	// Add a new column with the default value
+	_, err = conn.ExecContext(ctx, fmt.Sprintf("ALTER TABLE %s ADD COLUMN %s %s DEFAULT %s",
+		pq.QuoteIdentifier(cNewTableName),
+		pq.QuoteIdentifier(cNewColumnName),
+		columnType,
+		defaultExpr))
+	if err != nil {
+		return false, fmt.Errorf("failed to add column to schema-only copy: %w", err)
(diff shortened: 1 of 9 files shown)


internal/defaults/fastpath.go
```diff
@@ -0,0 +1,74 @@
...
+// adding a new column with a `DEFAULT` value of [defaultExpr] and checking
+// system catalogs to see if the fast-path optimization was applied.
+func UsesFastPath(ctx context.Context, conn db.DB, tableName, columnType, defaultExpr string) (bool, error) {
+	// Check if we have a real connection or a fake one
+	if _, ok := conn.(*db.FakeDB); ok {
+		return true, nil
+	}
+
+	// Create a schema-only copy of the table
+	_, err := conn.ExecContext(ctx, fmt.Sprintf("CREATE UNLOGGED TABLE %s AS SELECT * FROM %s WHERE false",
```
kvch, line 32: Why do we need a schema-only copy? What does the existing schema have to do with determining if the new column contains a volatile or non-volatile default?

andrew-farries: We need a table to add the column to in order to determine whether or not the default expression for the new column is volatile or not.

This is because there is currently no way in Postgres to determine volatility of a column default for a new column without actually creating the column. The procedure is:

* Create a table
* Add the column with the default to the table
* Read the system catalog to determine whether the column default was able to use the 'fast path' optimization or not.

If the fast path optimization was used to set the column default, then we know the default expression was non-volatile; otherwise its volatile.

We may be able to use an empty table with no columns rather than a schema-only copy of the actual table the column will be applied to. I'm not sure if there are circumstances under which that would make a difference; some indirect way that the column default expression could reference other columns in the same table perhaps?

kvch: Right, I understood the process. My questions is rather why we need the schema. Wouldn't it be simpler to just create an empty table with no columns?

pkg/migrations/op_add_column.go
```diff
@@ -247,7 +276,7 @@ func (o *OpAddColumn) Validate(ctx context.Context, s *schema.Schema) error {
 	return nil
 }
 
-func addColumn(ctx context.Context, conn db.DB, o OpAddColumn, t *schema.Table, tr SQLTransformer) error {
+func addColumn(ctx context.Context, conn db.DB, o OpAddColumn, t *schema.Table, fastPathDefault bool, tr SQLTransformer) error {
```
kvch, line 279: More of a general note that we should address in a later PR is that we could make `ColumnSQLWriter` a bit smarter and handle `fastPathDefault` automatically and all the other intricate column adding methods.

kvch (approved):

PR #697, opened 2025-02-21
Improve `create_index` operation with extra index element settings

kvch:
This PR adds a breaking change to `create_index` operation. From now on `columns` does not accept a list of column names. Instead it expects a map, so we can support for additional index elements settings.

New settings:
* `collation`: name of the collation to use on the column
* `opclass`: `opclass.name` sets the name of the operator class and `opclass.params` expects a KV pairs for operator class settings
* `sort`: change the ordering (`ASC` or `DESC`)
* `nulls`: change the position of null values in the index (`FIRST` or `LAST`)

### DIFF
--- a/docs/operations/create_index.mdx
+++ b/docs/operations/create_index.mdx
@@ -17,10 +17,10 @@ description: A create index operation creates a new index on a set of columns.
         "nulls": "FIRST | LAST",
         "opclass": {
           "name": "operator_class_name",
-          "params": {
-             "param1": "val",
-             "param2": "val"
-          }
+          "params": [
+             "param1=val",
+             "param2=val"
+          ]
         }
       }
     ]
--- a/docs/operations/create_index.mdx
+++ b/docs/operations/create_index.mdx
@@ -10,7 +10,20 @@ description: A create index operation creates a new index on a set of columns.
   "create_index": {
     "table": "name of table on which to define the index",
     "name": "index name",
-    "columns": [ "names of columns on which to define the index" ]
+    "columns": [
+      "column_name": {
+        "collate": "collation name",
+        "sort": "ASC | DESC",
+        "nulls": "FIRST | LAST",
+        "opclass": {
+          "name": "operator_class_name",
+          "params": {
+             "param1": "val",
+             "param2": "val"
+          }
+        }
+      }
+    ]
     "predicate": "conditional expression for defining a partial index",
     "storage_parameters": "comma-separated list of storage parameters",
     "unique": true | false,
@@ -51,3 +64,9 @@ Set storage parameters and index method:
 Create a unique index:
 
 <ExampleSnippet example="42_create_unique_index.json" language="json" />
+
+### Create an index with custom operator class
+
+Create an index with a custom operator class:
+
+<ExampleSnippet example="54_create_index_with_opclass.json" language="json" />
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.35.0
 	github.com/testcontainers/testcontainers-go/modules/postgres v0.35.0
-	github.com/xataio/pg_query_go/v6 v6.0.0-20250124115938-4fa82ad6d036
+	github.com/xataio/pg_query_go/v6 v6.0.0-20250226124908-814e464c6798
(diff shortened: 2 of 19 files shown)


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/stretchr/testify v1.10.0
 	github.com/testcontainers/testcontainers-go v0.35.0
 	github.com/testcontainers/testcontainers-go/modules/postgres v0.35.0
-	github.com/xataio/pg_query_go/v6 v6.0.0-20250124115938-4fa82ad6d036
+	github.com/xataio/pg_query_go/v6 v6.0.0-20250226124908-814e464c6798
```
kvch, line 18: It has to be updated after my PR is merged in pg_query_go: https://github.com/xataio/pg_query_go/pull/9

kvch:
This PR is ready for review, but `go.mod` needs to be updated before merge.

go.mod
kvch: Updated.

pkg/sql2pgroll/expect/create_index.go
```diff
@@ -21,46 +21,108 @@ func CreateIndexOp1WithMethod(method string) *migrations.OpCreateIndex {
...
+	Table:   "foo",
+	Method:  migrations.OpCreateIndexMethodBtree,
+	Columns: map[string]migrations.IndexElemSettings{"bar": {Nulls: ptr(migrations.IndexElemSettingsNullsLAST)}},
+}
+
+var CreateIndexOp10 = &migrations.OpCreateIndex{
+	Name:    "idx_name",
+	Table:   "foo",
+	Method:  migrations.OpCreateIndexMethodBtree,
+	Columns: map[string]migrations.IndexElemSettings{"bar": {Opclass: ptr(migrations.IndexElemSettingsOpclass{Name: "opclass", Params: map[string]any{"test": "test"}})}},
```
andrew-farries, line 91: Can break this up into multiple lines

pkg/sql2pgroll/create_index.go
```diff
@@ -18,12 +19,62 @@ func convertCreateIndexStmt(stmt *pgq.IndexStmt) (migrations.Operations, error)
...
+				if err != nil {
+					return nil, nil
+				}
+				for _, opt := range strings.Split(opts[1:len(opts)-1], ",") {
+					optKV := strings.SplitN(opt, "=", 2)
+					if len(optKV) != 2 {
+						continue
+					}
+					opclassOpts[optKV[0]] = optKV[1]
+				}
```
andrew-farries, line 49: Do we need to store these opclass options in a map?

Could the format be:

```json
{
  "name": "54_create_index_with_opclass",
  "operations": [
    {
      "create_index": {
        "name": "idx_fruits_custom_opclass",
        "table": "fruits",
        "columns": {
          "id": {
            "opclass": {
              "name": "int8_ops"
               "params": "foo = bar"
            }
          }
        }
      }
    }
  ]
}
```

Then we wouldn't need this manual deparsing. I don't think we use the extra structure of having the opclass options as a map.

pkg/migrations/op_create_index.go
```diff
@@ -34,9 +34,33 @@ func (o *OpCreateIndex) Start(ctx context.Context, conn db.DB, latestSchema stri
 		stmt += fmt.Sprintf(" USING %s", string(o.Method))
 	}
 
-	stmt += fmt.Sprintf(" (%s)", strings.Join(
-		quoteColumnNames(table.PhysicalColumnNamesFor(o.Columns...)), ", "),
-	)
+	if len(o.Columns) != 0 {
```
andrew-farries, line 37: Could we add comments to the contents of this block? Just one line above each block saying what each section is responsible for deparsing.

docs/operations/create_index.mdx
```diff
@@ -10,7 +10,20 @@ description: A create index operation creates a new index on a set of columns.
...
+      "column_name": {
+        "collate": "collation name",
+        "sort": "ASC | DESC",
+        "nulls": "FIRST | LAST",
+        "opclass": {
+          "name": "operator_class_name",
+          "params": {
+             "param1": "val",
+             "param2": "val"
+          }
```
andrew-farries, line 23: I think this doc is out of date now that `params` was changed to a slice of strings rather than a map.

andrew-farries (approved):

PR #709, opened 2025-03-04
Introduce `DBAction` interface and refactoring of column dropping

kvch:
This PR is part of a refactoring effort. My goal is to unify all database actions pgroll is executing. Example actions include dropping columns, renaming columns, creating and dropping triggers, creating constraints, etc.

We can define a list of these actions in a migration phase and run them sequentially. The following example is a `Complete` phase of an imaginary migration:

```golang
func Complete(ctx context.Context) error {
    actions := []DBAction{
        NewDropColumnAction(conn, tableName, columnName),
        NewRenameColumnAction(conn, tableName, newName, oldName),
        NewDropColumnAction(conn, tableName, CNeedsBackfill),
        NewDropTriggerAction(conn, triggerName),
    }

    for _, action := actions {
       if err := action.Execute(ctx); err != nil {
           return err
       }
    }
}
```

In this step I implemented column dropping as an action. In later PRs I would like to add creating and dropping triggers, functions, etc.

### DIFF
--- a/pkg/migrations/dbactions.go
+++ b/pkg/migrations/dbactions.go
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: Apache-2.0
+
 package migrations
 
 import (
--- a/pkg/migrations/dbactions.go
+++ b/pkg/migrations/dbactions.go
@@ -1,0 +1,47 @@
+package migrations
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/db"
+)
+
+// DBAction is an interface for common database actions
+// pgroll runs during migrations.
+type DBAction interface {
+	Execute(context.Context) error
+}
+
+// dropColumnAction is a DBAction that drops one or more columns from a table.
+type dropColumnAction struct {
+	conn db.DB
+
+	table   string
+	columns []string
+}
+
+func NewDropColumnAction(conn db.DB, table string, columns ...string) *dropColumnAction {
+	return &dropColumnAction{
+		conn:    conn,
+		table:   table,
+		columns: columns,
+	}
+}
+
+func (a *dropColumnAction) Execute(ctx context.Context) error {
+	_, err := a.conn.ExecContext(ctx, fmt.Sprintf("ALTER TABLE IF EXISTS %s %s",
+		pq.QuoteIdentifier(a.table),
+		a.dropMultipleColumns()))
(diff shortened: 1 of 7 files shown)


exekias:
@kvch code change LGTM but I would wait on the rest of the team to see if this is the desired path! my personal view:

I kind of like that operations are self-contained for the most part, with some helper functions for complex stuff like backfills. This pattern would force us to write the statements elsewhere, hiding the operation implementation a bit. Not a big deal for me, but still...

kvch:
The operations are getting less and less self-contained, because we are adding more and more complexity as go beyond the naive implementation of each migration. Also, some of the operations have overlap like `create_constraint` and `create_table` or adding a unique constraint and adding a unique index. Just to list a few examples.

Furthermore, we already have `createUniqueIndexConcurrently`, `alterSequenceOwner`, `RenameColumn` etc. These are already actions, and we are organically going towards this architecture. But instead of building an "accidental" architecture, I would like to preserve the clarity of `pgroll` and consciously refactor the code. The operations we perform in each migration phase fits perfectly the [command design pattern](https://en.wikipedia.org/wiki/Command_pattern). This is my attempt to create reusable "commands"/"actions" without the fluff. Thus, we can focus on putting together a migration from existing building blocks, instead of concatenating the same SQL strings in multiple operations.

exekias:
👍 thank you for the detailed explanation! no strong opinion on my side, to me whatever @andrew-farries and you decide works for me!

pkg/migrations/dbactions.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/lib/pq"
+	"github.com/xataio/pgroll/pkg/db"
+)
+
+// DBAction is an interface for common database actions
+// pgroll runs during migrations.
+type DBAction interface {
+	Execute(context.Context) error
+}
```
andrew-farries, line 18: We should consider moving this interface and implementations into a separate `dbactions` package IMO.

kvch: I agree, I will move it in a future PR. My initial goal with this PR was to get your feedback on the idea.

kvch:
I like this approach. To keep the review burden and the changeset small, I would like to transform all existing database actions to the common format in different PRs. Then in a separate PR I will implement the new interface for `Start`, `Complete` and `Rollback`. Do that work for you?

andrew-farries:
That works for me 👍

andrew-farries (approved):

PR #741, opened 2025-03-24
Do not run tests against PostgreSQL 14

kvch:
Remove jobs that run against PostgreSQL 14. We agreed to support the last 3 version
of PostgreSQL. The lastest version of PostgreSQL is 17. So the last 3 versions we need to support is 15, 16 and 17.

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -73,22 +73,18 @@ branches:
 
       # Required. The list of status checks to require in order to merge into this branch
       contexts:
-      - test (pg: 14.8, schema: public)
       - test (pg: 15.3, schema: public)
       - test (pg: 16.0, schema: public)
       - test (pg: 17.0, schema: public)
       - test (pg: latest, schema: public)
-      - test (pg: 14.8, schema: non_public)
       - test (pg: 15.3, schema: non_public)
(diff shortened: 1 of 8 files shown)


andrew-farries (commented):
Some more things that need to change if we drop PG14 support:

* The conditional creation of views with `(security_invoker=true)`:
 https://github.com/xataio/pgroll/blob/c15326b533bb26b4ff84f600bd7dd446ece66c66/pkg/roll/execute.go?plain=1#L294-L303

and

https://github.com/xataio/pgroll/blob/c15326b533bb26b4ff84f600bd7dd446ece66c66/pkg/roll/execute_test.go?plain=1#L430-L432

* `README.md` sections:

https://github.com/xataio/pgroll/blob/c15326b533bb26b4ff84f600bd7dd446ece66c66/docs/getting-started.md?plain=1#L85-L89

and 

https://github.com/xataio/pgroll/blob/c15326b533bb26b4ff84f600bd7dd446ece66c66/README.md?plain=1#L19-L28

andrew-farries (approved):

kvch:
We agreed to keep support for PG14 until its end of life. Ref: https://endoflife.date/postgresql

PR #746, opened 2025-03-25
Implement rename column actions as `DBAction`

kvch:
This PR implements a new `DBAction` for renaming columns. The action expects a table name, "source" column name and a "target" column name.


pkg/migrations/op_add_column.go
```diff
@@ -102,11 +102,8 @@ func (o *OpAddColumn) Start(ctx context.Context, conn db.DB, latestSchema string
 }
 
 func (o *OpAddColumn) Complete(ctx context.Context, conn db.DB, s *schema.Schema) error {
-	_, err := conn.ExecContext(ctx, fmt.Sprintf("ALTER TABLE IF EXISTS %s RENAME COLUMN %s TO %s",
-		pq.QuoteIdentifier(o.Table),
-		pq.QuoteIdentifier(TemporaryName(o.Column.Name)),
-		pq.QuoteIdentifier(o.Column.Name),
-	))
+	renameColumn := NewRenameColumnAction(conn, o.Table, TemporaryName(o.Column.Name), o.Column.Name)
+	err := renameColumn.Execute(ctx)
```
andrew-farries, line 106: Would it be better to combine these steps:

```go
err := NewRenameColumnAction(conn, o.Table, TemporaryName(o.Column.Name), o.Column.Name).Execute(ctx)
```

here and for other actions?

andrew-farries (approved):

PR #747, opened 2025-03-25
Implement creating a unique index concurrently as a `DBAction`

kvch:
This PR implements creating unique indexes concurrently as a `DBAction`. Looking at the implementation of the action, we can refactor it in a later PR and introduce a `CreateIndexSQLWriter` similar to `ConstraintSQLWriter`. But this is a problem for a future PR.


andrew-farries (approved):
IIRC, we don't have automated tests to check https://github.com/xataio/pgroll/issues/485.

It would be worth running through one of the reproductions there to ensure that creating indexes under load is still fixed after this extraction.

kvch:
I tested the scenario in the issue mentioned. The issue is not coming back when we merge this change.

PR #748, opened 2025-03-25
Use `license-header-checker` locally to add missing license headers

kvch:
Copilot removes the license headers from some files. I added a make target for adding the license headers back to these files. Or just to add it to new files.

The target uses the same binary we leverage in GH actions.


andrew-farries (commented):
There is an empty file; `add_license_header.sh` committed as part of this PR that should be removed.

We should update the GH actions workflow so that it runs `make license` too rather than duplicating the same steps.

andrew-farries (approved):

PR #749, opened 2025-03-25
Implement creating triggers as `DBAction`

kvch:
This PR implements a new `DBAction` for creating triggers.

### DIFF
--- a/pkg/migrations/op_add_column.go
+++ b/pkg/migrations/op_add_column.go
@@ -78,8 +78,7 @@ func (o *OpAddColumn) Start(ctx context.Context, conn db.DB, latestSchema string
 
 	var tableToBackfill *schema.Table
 	if o.Up != "" {
-		createUpTrigger := NewCreateTriggerAction(
-			conn,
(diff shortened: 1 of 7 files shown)


pkg/migrations/op_add_column.go
```diff
@@ -78,16 +78,20 @@ func (o *OpAddColumn) Start(ctx context.Context, conn db.DB, latestSchema string
...
-			Direction:      TriggerDirectionUp,
-			Columns:        table.Columns,
-			SchemaName:     s.Name,
-			LatestSchema:   latestSchema,
-			TableName:      table.Name,
-			PhysicalColumn: TemporaryName(o.Column.Name),
-			SQL:            o.Up,
-		})
+		createUpTrigger := NewCreateTriggerAction(
+			conn,
```
andrew-farries, line 82: nit: Could put `conn` on the same line as `NewCreateTriggerAction` here and everywhere else; makes things a bit more readable IMO.

andrew-farries (approved):

PR #755, opened 2025-03-26
Initialize YAML migration file support

kvch:
This PR adds basic support for reading YAML migration files. From now on users can read migrations from YAML files as well. A migration directory can store both JSON and YAML files at the same time. Ordering of the files is based on the filename without the file extension.

The following migration folder content is valid:
```sh
ls migrations
01_create_table.json
02_create_index.yaml
03_create_constraint.json
```

Also, function that are responsible for reading migration files are ready to handle SQL migrations as well.

This is not the complete support/feature, the PR has several limitations:
* `pgroll` reads YAML migration files, but when the user tried to get the migration history `pgroll` return JSON migrations.
* Documentation does not show YAML examples

### DIFF
--- a/pkg/migrations/op_common.go
+++ b/pkg/migrations/op_common.go
@@ -11,7 +11,6 @@ import (
 	"maps"
 	"path/filepath"
 	"slices"
-	"sort"
 	"strings"
 
 	"gopkg.in/yaml.v3"


andrew-farries (commented):
Looks great 👍 Can't wait to start writing YAML migrations 🚀

andrew-farries (approved):

Issue #726, opened 2025-03-13
`pgroll latest --local` needs db access?


richardgill:
```
❯ maki roll latest --local pgroll
Error: Error: dial tcp [::1]:5432: connect: connection refused
 (id: undefined)
```

andrew-farries:
Currently, `pgroll latest --local` needs db access is to support the `--with-schema` flag:

```
$ ls migrations/
01_create_tables.json
02_create_another_table.json
```

```
$ pgroll latest --local migrations/ --with-schema
public_02_create_another_table
```

The database connection is required in order to know which schema (`public` in this case) `pgroll` is set to work with.

andrew-farries:
This should be resolved by https://github.com/xataio/pgroll/pull/760 so that `pgroll latest --local` does not require any connection to the database.

PR #760, opened 2025-03-27
Dont require database access to `pgroll latest --local`

andrew-farries:
Ensure that the `latest` command does not require database access when the `--local` flag is provided. 

In this case the command should act only locally, with the migrations on disk and not require a connection to the database.

The reason it used to require a DB connection was to support the `--with-schema` flag which took the schema name from an instantiated `Roll` instance. Change this so that the schema name is instead taken from the command line `--schema` flag (or the `PGROLL_SCHEMA` env var) directly instead.

Fixes https://github.com/xataio/pgroll/issues/726

### DIFF
--- a/cmd/latest.go
+++ b/cmd/latest.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	"github.com/spf13/cobra"
+	"github.com/xataio/pgroll/pkg/roll"
 )
 
 func latestCmd() *cobra.Command {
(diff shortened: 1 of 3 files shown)


andrew-farries:
rebased to resolve a merge conflict

cmd/latest.go
```diff
@@ -21,48 +24,21 @@ func latestCmd() *cobra.Command {
...
 				}
 			}
 
-			var prefix string
-			if withSchema {
-				prefix = m.Schema() + "_"
-			}
-
-			fmt.Printf("%s%s\n", prefix, latestVersion)
+			fmt.Printf("%s\n", latestVersion)
```
kvch, line 41:
```suggestion
			fmt.Println(latestVersion)
```

kvch (approved):
Smallest nit of my life

PR #772, opened 2025-04-01
Replace JSON examples with YAML && use them in docs

kvch:
This PR converts JSON migrations files in `examples` folder to YAML files. The references to the JSON files in the documentation are replaced with references to YAML migration files.

Requires: https://github.com/xataio/pgroll/pull/771


exekias:
@kvch I thought we wanted to support YAML as a second format, but not drop JSON support. Do we want to switch all examples to yaml?

kvch:
We are not dropping support for JSON. We agreed to change all examples to YAML, and display both YAML and JSON format in the documentations. But we do not need to keep both formats, because we can convert YAML to JSON in the docs automatically.

exekias (approved):

PR #779, opened 2025-04-04
Add fallback method for backfilling records

kvch:
Previously, `pgroll` could not backfill tables that did not have a primary key or a unique column. From now on, `pgroll` can fallback to use the `_pgroll_needs_backfill` internal column to select the next batch to backfill.

Required by #698

### DIFF
--- a/pkg/backfill/backfill.go
+++ b/pkg/backfill/backfill.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/lib/pq"
 	"github.com/xataio/pgroll/pkg/backfill/templates"
 	"github.com/xataio/pgroll/pkg/db"
 	"github.com/xataio/pgroll/pkg/schema"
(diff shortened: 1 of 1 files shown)


pkg/backfill/backfill.go
```diff
@@ -9,11 +9,14 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/lib/pq"
 	"github.com/xataio/pgroll/pkg/backfill/templates"
 	"github.com/xataio/pgroll/pkg/db"
 	"github.com/xataio/pgroll/pkg/schema"
 )
 
+const CNeedsBackfillColumn = "_pgroll_needs_backfill"
```
exekias, line 18: a comment here would help

exekias (approved):
great addition! would it be possible to add some tests for this case?

kvch:
I added one more test and a comment.

PR #780, opened 2025-04-04
Remove the unnecessary file parameter from the help text

### LINKED ISSUE #776: rollback command help indicates accepting a file
I installed pgroll via Homebrew:
```
$ pgroll --version
pgroll version 0.10.0
```

The inline help indicates accepting a file
```
$ pgroll rollback --help
Roll back an ongoing migration

Usage:
  pgroll rollback <file> [flags]
...
```

I suppose it shouldn't, the online documentation doesn't mention it: https://pgroll.com/docs/latest/cli/rollback

kvch:
Closes #776

### DIFF
--- a/cmd/rollback.go
+++ b/cmd/rollback.go
@@ -10,7 +10,7 @@ import (
 )
 
 var rollbackCmd = &cobra.Command{
-	Use:   "rollback <file>",
+	Use:   "rollback",
 	Short: "Roll back an ongoing migration",
 	RunE: func(cmd *cobra.Command, args []string) error {
 		m, err := NewRoll(cmd.Context())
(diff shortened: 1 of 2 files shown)


cmd/rollback.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 var rollbackCmd = &cobra.Command{
-	Use:   "rollback <file>",
+	Use:   "rollback",
```
gulcin, line 13: Do we need to change the help text as well?

kvch: You mean the generated cli def? I've just pushed that change. Otherwise this is the help text we display.

gulcin: Ah yes, now seen both files, thank you.

gulcin (approved):

PR #781, opened 2025-04-07
Improve error message when `start` and `rollack` fails

kvch:
When `pgroll` fails to execute `start` phase of the migration, it runs `rollback` immediately. If `rollback` fails as well, the errors are joined. This made the error message hard to understand.

Previously, `pgroll` emitted this message:
```
Error: unable to execute start operation: pq: relation "qsdf" does not exist
pq: current transaction is aborted, commands ignored until end of transaction block
```

From now on, the message is clearer.
```
Error: unable to execute start operation of 'my_migration_name': pq: relation "qsdf" does not exist
unable to roll back failed operation: pq: current transaction is aborted, commands ignored until end of transaction block
```

If `start` fails but rollback succeeds, we emit a clearer error message:
```
Error: rolled back 'my_migration_name' because start command has failed: pq: relation "qsdf" does not exist
```

### DIFF
--- a/pkg/roll/execute.go
+++ b/pkg/roll/execute.go
@@ -89,10 +89,12 @@ func (m *Roll) StartDDLOperations(ctx context.Context, migration *migrations.Mig
 		table, err := op.Start(ctx, m.pgConn, latestSchema, newSchema)
 		if err != nil {
 			errRollback := m.Rollback(ctx)
-
-			return nil, errors.Join(
-				fmt.Errorf("unable to execute start operation: %w", err),
-				errRollback)
+			if errRollback != nil {
+				return nil, errors.Join(
+					fmt.Errorf("unable to execute start operation of %q: %w", migration.Name, err),
+					fmt.Errorf("unable to roll back failed operation: %w", errRollback))
+			}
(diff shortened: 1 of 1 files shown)


exekias (approved):

pkg/roll/execute.go
```diff
@@ -89,10 +89,12 @@ func (m *Roll) StartDDLOperations(ctx context.Context, migration *migrations.Mig
...
-
-			return nil, errors.Join(
-				fmt.Errorf("unable to execute start operation: %w", err),
-				errRollback)
+			if errRollback != nil {
+				return nil, errors.Join(
+					fmt.Errorf("unable to execute start operation of %q: %w", migration.Name, err),
+					fmt.Errorf("unable to roll back failed operation: %w", errRollback))
+			}
+			return nil, fmt.Errorf("rolled back %q because start command has failed: %w", migration.Name, err)
```
exekias, line 97: not sure if it improves it, feel free to disregard!

```suggestion
			return nil, fmt.Errorf("failed to start %q migration, changes rolled back: %w", migration.Name, err)
```

PR #788, opened 2025-04-10
Pass all column attributes to new temporary column when adding a new one to the table

kvch:
The following migration was failing due to several reasons:

```json
{
  "operations": [
    {
      "create_table": {
        "name": "users",
        "columns": [
          {
            "name": "id",
            "type": "uuid",
            "pk": true
          }
        ]
      }
    },
    {
      "add_column": {
        "table": "users",
        "column": {
          "name": "name",
          "type": "text",
          "nullable": true
        }
      }
    },
    {
      "create_constraint": {
        "table": "users",
        "columns": [
          "name"
        ],
        "type": "unique",
        "name": "my_custom_name",
        "up": {
          "name": "name"
        },
        "down": {
          "name": "name"
        }
      }
    }
  ]
}
```

This PR fixes the first problem with the migration. The first problem is when `pgroll` adds the new column with the temporary name to the table schema it lacks all column attributes. Thus, if a later operation tries to duplicate this duplicated column it fails because we do not know the type of the column, the default value, etc. The SQL statement generated has a syntax error because of the lack of information:

```
Failed to start migration: unable to execute start operation: failed to duplicate column: pq: syntax error at or near ","
```

### DIFF
--- a/pkg/migrations/op_add_column.go
+++ b/pkg/migrations/op_add_column.go
@@ -99,9 +99,17 @@ func (o *OpAddColumn) Start(ctx context.Context, conn db.DB, latestSchema string
 		tableToBackfill = table
 	}
 
-	table.AddColumn(o.Column.Name, &schema.Column{
-		Name: TemporaryName(o.Column.Name),
-	})
+	tmpColumn := &schema.Column{
+		Name:     TemporaryName(o.Column.Name),
+		Type:     o.Column.Type,
+		Default:  o.Column.Default,
+		Nullable: o.Column.Nullable,
+		Unique:   o.Column.Unique,
+	}
+	if o.Column.Comment != nil {
+		tmpColumn.Comment = *o.Column.Comment
+	}
+	table.AddColumn(o.Column.Name, tmpColumn)
 
 	return tableToBackfill, nil
 }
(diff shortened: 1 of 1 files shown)


exekias:
would it be possible to add a test that fails without this change?

exekias:
Do I understand correctly that you say it is not fixing anything either?

kvch:
It is fixing the syntax error. If we fix the syntax error, then we get to the underlying issue which is that we cannot duplicate the column.

pkg/migrations/op_add_column.go
```diff
@@ -99,9 +99,17 @@ func (o *OpAddColumn) Start(ctx context.Context, conn db.DB, latestSchema string
 		tableToBackfill = table
 	}
 
-	table.AddColumn(o.Column.Name, &schema.Column{
-		Name: TemporaryName(o.Column.Name),
-	})
+	tmpColumn := &schema.Column{
```
exekias, line 102: to avoid mistakes if a new field is added to `schema.Column` we probably want to clone `o.Column` then change the name maybe?

kvch: Ok, I am refactoring the code.

kvch: These are two different `Column` types, I cannot really clone them. I added a function we could call to translate from `migrations.Column` to `schema.Column`.

PR #795, opened 2025-04-17
Add support for running SQL migration files using `pgroll`

kvch:


kvch:
Closing because it is outdated.

PR #800, opened 2025-04-25
Do not display postgresql options in `convert` help text

### LINKED ISSUE #716: `pgroll convert --help` wrong
These are not options for convert? (or are they?)

Image

kvch:
Do not list global PostgreSQL options in help text. Previously the help text contained postgresql options:

```sh
$ go run . convert --help
Convert SQL statements to a pgroll migration. The command can read SQL statements from stdin or a file

Usage:
  pgroll convert <path to file with migrations> [flags]

Flags:
  -h, --help          help for convert
  -n, --name string   Name of the migration

Global Flags:
      --lock-timeout int       Postgres lock timeout in milliseconds for pgroll DDL operations (default 500)
      --pgroll-schema string   Postgres schema to use for pgroll internal state (default "pgroll")
      --postgres-url string    Postgres URL (default "postgres://postgres:postgres@localhost?sslmode=disable")
      --role string            Optional postgres role to set when executing migrations
      --schema string          Postgres schema to use for the migration (default "public")

```
From now these "global" flags do not show up:

```sh
$ go run . convert --help
Convert SQL statements to a pgroll migration. The command can read SQL statements from stdin or a file

Usage:
  pgroll convert <path to file with migrations> [flags]

Flags:
  -h, --help          help for convert
  -n, --name string   Name of the migration
```

Closes #716


cmd/root.go
```diff
@@ -47,25 +47,13 @@ func Prepare() *cobra.Command {
...
-	rootCmd.PersistentFlags().String("schema", "public", "Postgres schema to use for the migration")
-	rootCmd.PersistentFlags().String("pgroll-schema", "pgroll", "Postgres schema to use for pgroll internal state")
-	rootCmd.PersistentFlags().Int("lock-timeout", 500, "Postgres lock timeout in milliseconds for pgroll DDL operations")
-	rootCmd.PersistentFlags().String("role", "", "Optional postgres role to set when executing migrations")
-
-	viper.BindPFlag("PG_URL", rootCmd.PersistentFlags().Lookup("postgres-url"))
-	viper.BindPFlag("SCHEMA", rootCmd.PersistentFlags().Lookup("schema"))
-	viper.BindPFlag("STATE_SCHEMA", rootCmd.PersistentFlags().Lookup("pgroll-schema"))
-	viper.BindPFlag("LOCK_TIMEOUT", rootCmd.PersistentFlags().Lookup("lock-timeout"))
-	viper.BindPFlag("ROLE", rootCmd.PersistentFlags().Lookup("role"))
```
andrew-farries, line 60: This removes the ability to set these flags via env vars, which I don't think we want.

andrew-farries (commented):
I'm not sure we need to do this - the removed flags are listed under 'Global flags' and I don't think there is an expectation in other CLI tools that they must apply to **all** subcommands, just most of them.

kvch:
Ok, then I am closing the issue as wontfix.

PR #801, opened 2025-04-28
Remove `name` field from all examples

andrew-farries:
Remove the `name` field from all example migrations.

Migration names now default to the file name if a `name` field is not present and we recommend that approach over explicitly naming the migration in the JSON/YAML.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -90,7 +90,7 @@ operations:
 
 Take this file and save it as `sql/01_create_users_table.yaml`.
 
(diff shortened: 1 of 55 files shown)


kvch (approved):
The changes look good. Do you want to remove `name` from other documentation pages as well like in the tutorial page?

andrew-farries:
> The changes look good. Do you want to remove `name` from other documentation pages as well like in the tutorial page?

It's a good point - I'll go through the docs and remove `name` fields there too.

PR #804, opened 2025-04-29
Add verbose logging

### LINKED ISSUE #784: Verbose mode
We would like to add a new flag to `start`, `complete` and `rollback` named `verbose`. When this flag is set, `pgroll` prints what operations it is running to show users what is going on exactly.

Example output for `start`:
```
$ pgroll start --verbose
Starting migration my_migration
Executing operation #1: create_table
New table created: my_table
Executing operation #2: create_index
New index created: my_index on my_table
Executing operation #3: add_column
New column added: my_new_column
New schema is ready public_my_migration
Backfill process has started
Backfill process is done
```

Example output for `rollback`:
```
$ pgroll rollback --verbose
Rolling back migration my_migration
Rolling back operation #3: add_column
New column dropped: my_new_column
Rolling back operation #2: create_index
Index dropped: my_index on my_table
Rolling back operation #1: create_table
Table dropped: my_table
Schema public_my_migration is dropped
```

kvch:
This PR adds a new flag called `--verbose`. If you set it, `pgroll` will be more verbose.
It logs the following events with contextual info to stdout:

* starting migration
* starting operation
* starting backfilling
* completing operation
* completing migration
* completing backfilling
* creating versioned schema
* dropping versioned schema

Example output:
```

```

Closes #784


andrew-farries (commented):
This works nicely as a CLI tool, but couples us tightly to the `pterm.Logger` type which will be less suitable when using `pgroll` as a dependency from other Go code. I think it's fine for now but we may want to revisit this.

A general suggestion would be to introduce an interface something like:

```go
  type OperationLogger interface {
      // Operation lifecycle methods
      LogOperationStart(op Operation)
      LogOperationComplete(op Operation)
      LogOperationRollback(op Operation)

      // Specific logging events for common scenarios
      LogBackfillStart(table string)
      LogBackfillComplete(table string)
      LogConstraintValidation(table, constraint string)
      LogSchemaCreation(schemaName string)

      // Generic methods for custom events
      LogInfo(msg string, args ...any)
      LogError(msg string, err error, args ...any)
  }
```

and implement it for `pterm` such that the different operations can do:

```go
      logger.LogOperationStart(o)
```

rather than having to repeat themselves with "starting operation..." etc everywhere.

The `loggerArgs` methods could then move to the implementation of the `OperationLogger` interface:

```go
func (l *ptermOperationLogger) extractOperationArgs(op Operation) []any {
      // Extract common fields
      var args []any

      // Type-switch to extract operation-specific fields
      switch o := op.(type) {
      case *OpAddColumn:
          args = append(args,
              "operation", OpNameAddColumn,
              "name", o.Column.Name,
              "type", o.Column.Type,
              "table", o.Table,
              "nullable", o.Column.Nullable,
              "unique", o.Column.Unique,
          )
      case *OpCreateTable:
          args = append(args,
              "operation", OpNameCreateTable,
              "name", o.Name,
          )
          if len(o.Columns) > 0 {
              columnNames := getColumnNames(o.Columns)
              args = append(args, "columns", columnNames)
          }
      // Handle other operation types
      }

      return args
  }
```

this would keep the operations themselves relatively free of logging concerns. WDYT?

andrew-farries (approved):

PR #809, opened 2025-04-29
Restructure and rewrite the front page of the docs

kvch:


docs/getting-started.md
```diff
@@ -1,6 +1,31 @@
-# Getting started
+# Welcome to pgroll's documentation
 
-This guide covers the key concepts of `pgroll` and how to get started with it.
+`pgroll` ensures 100% uptime during schema changes in your application. It is an open source command line interface for developers using [PostgreSQL](https://postgresql.org) as their data layer.
```
andrew-farries, line 3: I think the intro section should be a bit longer, something  like this:

```suggestion
`pgroll` is a command-line tool for managing zero-downtime, reversible schema migrations in PostgreSQL. It follows an expand/contract workflow, using views to allow multiple versions of an application to work with the database during a migration. This enables safe, incremental changes—like adding columns or changing data types—without locking tables or disrupting service. `pgroll` also makes it easy to roll back changes if needed, making schema evolution safer and more reliable in production environments.```

File: docs/getting-started.md [hunk 2]
```diff
@@ -1,6 +1,31 @@
...
+
+## Getting started
+
+Learn how `pgroll` simplifies zero-downtime migrations for you
+
+* [Why use `pgroll`](why-use-pgroll)
```
andrew-farries, line 9: Inline code in list items looks odd (at least in the preview):

image

I think we don't need to use backticks here.

```suggestion
* [Why use pgroll](why-use-pgroll)
```

docs/getting-started.md
```diff
@@ -1,6 +1,31 @@
...
+* [Installation](installation)
+* [Concepts](concepts)
+* [Write your first migration](tutorial)
+
+## Guides
+
+Learn how to use `pgroll` in your developer workflow
+
+* [Integrate `pgroll` into your project](guides/clientapps)
```
andrew-farries, line 18:
```suggestion
* [Integrate pgroll into your project](guides/clientapps)
```

docs/getting-started.md
kvch: The intro section is short on purpose, because the goal of this page is to provide an outline for the whole documentation. What you are suggesting could go into the "why use pgroll" page.

andrew-farries: we could keep it shorter with something like:

```suggestion
`pgroll` is a migration tool for PostgreSQL that makes it easy to apply schema changes safely, without application downtime. It supports instant rollbacks and helps ensure applications stay available during database schema changes by allowing applications to access different versions of your database schema.
```

In particular, I think we should avoid saying `pgroll` ensures 100% uptime during schema changes in your application` as there are a bunch of reasons not related to the database schema why that might not be true.

andrew-farries (approved):
Left one follow up comment [here](https://github.com/xataio/pgroll/pull/809/files#r2077857186), up to you what you want to do there.

kvch:
Thank you! I applied your suggestion.

PR #814, opened 2025-05-06
Add new section to docs sidebar: Release notes

kvch:
This PR adds a new section to the left sidebar called: release notes. It contains links to past `pgroll` release notes on Github. It is more of an idea, than a thought. I have seen other documentations linking to past release notes. Maybe we can include such a page as well.

### DIFF
--- a/docs/config.json
+++ b/docs/config.json
@@ -221,6 +221,11 @@
           "file": "docs/operations/set_replica_identity.mdx"
         }
       ]
+    },
+    {
+      "title": "Release notes",
+      "href": "/releases",
+      "file": "docs/releases/README.md"
     }
   ]
 }
--- a/docs/releases/README.md
+++ b/docs/releases/README.md
@@ -1,0 +1,20 @@
+# Release notes
+
+Check out the release notes of past versions:
+
+* [v0.12.0](https://github.com/xataio/pgroll/releases/tag/v0.12.0)
+* [v0.11.1](https://github.com/xataio/pgroll/releases/tag/v0.11.1)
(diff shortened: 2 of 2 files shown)


kvch:
Converted to draft because the preview is not working yet.

kvch:
Fixed.

docs/releases/README.md
```diff
@@ -0,0 +1,20 @@
+# Release notes
+
+Check out the release notes of past versions:
+
+* [v0.12.0](https://github.com/xataio/pgroll/releases/tag/v0.12.0)
```
andrew-farries, line 5: Doesn't exist yet 😄

andrew-farries (commented):
If we want to do this then I think we also need a way to keep the page up to date as part of the release workflow, otherwise I think we'll forget to keep it up to date.

An alternative would be just to link to the Github releases page from elsewhere in the documentation.

docs/config.json
```diff
@@ -221,6 +221,11 @@
           "file": "docs/operations/set_replica_identity.mdx"
         }
       ]
+    },
+    {
+      "title": "Release notes",
+      "href": "/releases",
+      "file": "docs/releases/README.md"
```
andrew-farries, line 228: I'm not sure what causes it but this new page appears with this 404 link at the bottom of the page:

image

kvch: It happens on all previews for some reason. I did not want to bug Richard with it because it did not seem important enough. :D

kvch:
Yeah, I agree about maintaining it in the long run. I tried adding a link to the sidebar, but it does not seem the work with the current doc engine. Maybe it could be moved to the bottom of the front page after the details about connecting with us.

PR #821, opened 2025-05-09
Refactoring of writing migrations to file

kvch:
This PR adds a new type `MigrationWriter` to provied a unified way to write migrations
to file either in YAML or in JSON format. It is used in 3 commands: `convert`, `create` and `pull`.

Also with this change I added a new flag to `convert` called `json`. Previously, the command serialized migrations
into JSON format. From now on, it follows the same pattern like `create` and `pull`. It writes migrations in YAML format unless the user sets `--json` flag.

### DIFF
--- a/pkg/migrations/writer.go
+++ b/pkg/migrations/writer.go
@@ -1,0 +1,86 @@
+package migrations
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+
+	"sigs.k8s.io/yaml"
+)
+
+type MigrationFormat int
+
+const (
+	InvalidMigrationFormat MigrationFormat = iota
+	YAMLMigrationFormat
+	JSONMigrationFormat
+)
+
+var (
+	ErrInvalidMigrationFormat = errors.New("invalid migration format")
+)
+
+// MigrationWriter is responsible for writing Migrations and RawMigrations
+// to the configured io.Writer instance in either YAML or JSON.
+type MigrationWriter struct {
+	writer io.Writer
+	format MigrationFormat
+}
+
+// NewMigrationFormat returns YAML or JSON format
+func NewMigrationFormat(useJSON bool) MigrationFormat {
+	if useJSON {
+		return JSONMigrationFormat
+	}
+	return YAMLMigrationFormat
+}
+
+// Extension returns the extension name for the migration file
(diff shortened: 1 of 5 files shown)


pkg/migrations/writer.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (w *MigrationWriter) writeAny(migration any) error {
+	switch w.format {
+	case YAMLMigrationFormat:
+		yml, err := yaml.Marshal(migration)
+		if err != nil {
+			return err
+		}
+		_, err = w.writer.Write(yml)
+		return fmt.Errorf("encode yaml migration: %w", err)
```
andrew-farries, line 74: We're always returning an error here which is currently breaking all commands that try to write migrations as YAML.

pkg/migrations/writer.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"io"
+
+	"sigs.k8s.io/yaml"
+)
+
+type MigrationFormat int
+
+const (
+	InvalidMigrationFormat MigrationFormat = iota
```
andrew-farries, line 17: I don't think we need this. It's not used.

pkg/migrations/writer.go
```diff
@@ -0,0 +1,85 @@
...
+
+// NewWriter creates a new MigrationWriter
+func NewWriter(w io.Writer, f MigrationFormat) *MigrationWriter {
+	return &MigrationWriter{
+		writer: w,
+		format: f,
+	}
+}
+
+func (w *MigrationWriter) Write(m *Migration) error {
```
andrew-farries, line 58: Add a doc comment like the other exported methods.

pkg/migrations/writer.go
```diff
@@ -0,0 +1,85 @@
...
+	return w.writeAny(m)
+}
+
+func (w *MigrationWriter) WriteRaw(m *RawMigration) error {
```
andrew-farries, line 62: doc comment

andrew-farries (commented):
Nice change 👍 

It's currently broken though because marshalling YAML always returns an error 😛

qybhyv-4jecxy-Qastaw (approved):

qybhyv-4jecxy-Qastaw (approved):

andrew-farries (approved):

PR #825, opened 2025-05-12
Simplify README.md

kvch:
This PR simplifies the readme of `pgroll`.

I moved the section "How pgroll works" lower, before the documentation link. Now installation and basic tutorial are easier to find. It does not get lost in the 1000 lines long readme.

I also added two new badges: documentation and download count.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ See the [introductory blog post](https://pgroll.com/blog/introducing-pgroll-zero
 
 - [Installation](#installation)
 - [Usage](#usage)
+- [How pgroll works](#how-pgroll-works)
 - [Documentation](#documentation)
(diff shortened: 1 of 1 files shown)


exekias:
I believe this section is valuable for understanding what the project is about. I would be more ok with moving it down if we believe it's a problem. Personally I'm fine with it as it is

gulcin (changes requested):
I wonder if we can keep it but maybe place it somewhere after the necessary information you'd like to emphasize? I think @exekias suggested smth similar.

kvch:
I added the section about pgroll internals back to the readme.

gulcin (approved):
LGTM!

PR #827, opened 2025-05-12
Fail early in `migrate` if there is an incompatible migration file

### LINKED ISSUE #766: Fail early in case there is a breaking change in pgroll
In case there is a breaking change in `pgroll`, the CLI should detect it and fail as early as possible (so that it doesn't accidentally corrupt any state in the case of a breaking change). 

## Proposal 

Store the version of pgroll CLI with `pgroll init` and in the case of a breaking change, detect the version in database and fail early. 
(pointed out by Carlos) This is tricky in case there is a regression, `pgroll init` storing state might stop the user from downgrading the CLI. 

Potential solutions:
- we maintain a compatible versions map and up / downgrading works unless the version boundary is crossed. 
- we only show a warning and/or a confirmation dialog and let user override.

Related https://github.com/xataio/pgroll/issues/765

kvch:
This PR changes the `migrate` command. It does not start any migrations before it makes sure that the outstanding migration files are  compatible with the current version.

Closes #766

### DIFF
--- a/cmd/migrate.go
+++ b/cmd/migrate.go
@@ -80,7 +80,6 @@ func migrateCmd() *cobra.Command {
 
 			// fail early if there is an incompatible migration
 			migs, err := parseMigrations(rawMigs)
-			fmt.Println(migs)
 			if err != nil {
 				return fmt.Errorf("failed to run migrate: %w", err)
 			}
--- a/cmd/migrate.go
+++ b/cmd/migrate.go
@@ -80,6 +80,7 @@ func migrateCmd() *cobra.Command {
 
 			// fail early if there is an incompatible migration
 			migs, err := parseMigrations(rawMigs)
+			fmt.Println(migs)
 			if err != nil {
 				return fmt.Errorf("failed to run migrate: %w", err)
 			}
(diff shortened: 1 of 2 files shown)


andrew-farries (commented):
Looks good, one suggested change 👍

cmd/migrate.go
```diff
@@ -69,16 +68,23 @@ func migrateCmd() *cobra.Command {
...
 
-			if len(migs) == 0 {
+			if len(rawMigs) == 0 {
 				fmt.Println("Database is up to date; no migrations to apply")
 				return nil
 			}
 
+			// fail early if there is an incompatible migration
+			migs, err := parseMigrations(rawMigs)
+			fmt.Println(migs)
```
andrew-farries, line 83: Remove

cmd/migrate.go
```diff
@@ -104,11 +110,20 @@ func migrateCmd() *cobra.Command {
...
+		m, err := migrations.ParseMigration(rawMigration)
+		if err != nil {
+			errs = errors.Join(errs, err)
+		}
+		parsedMigrations = append(parsedMigrations, m)
 	}
-
-	return runMigration(ctx, m, parsedMigration, complete, c)
+	if errs != nil {
+		return nil, fmt.Errorf("incompatible migration(s) found: %w. please update the files.", errs)
```
andrew-farries, line 126: Avoid saying `update the files` here because this assumes the cause of the error. All we know here is that a migration couldn't be parsed; could be a typo.

```suggestion
				return nil, fmt.Errorf("incompatible migration(s): %w", errs)
```

Issue #364, opened 2024-06-27
Tooling for baseline migration on existing database


kostasb:
For the purpose of managing an already existing database with pgroll-managed, provide a tool that initiates a baseline migration and enables pgroll on the database.

PR #832, opened 2025-05-13
[baseline] Add `pgroll baseline` command

andrew-farries:
Add a new `pgroll baseline` command to `pgroll` that allows users to establish a baseline migration from an existing database schema.

### Usage

Run:

```bash
$ pgroll baseline 01_initial_version migrations/
```

This does two things:
* Creates a new entry in `pgroll.migrations` with a new migration type of `baseline`. The migration JSON is empty and the `resulting_schema` field contains the schema at the time the command was run.
* Creates a placeholder `migrations/01_initial_version.yaml` file for the user to populate (likely with the use of `pg_dump`).

### Features

  - New `pgroll baseline <version> <target directory>` command (currently hidden)
  - Creates a `baseline` migration in the `pgroll.migrations` table that captures the current schema state without applying changes.
  - Validation to prevent baseline creation during active migrations
  - Adds test coverage for baseline functionality

### Implementation Details

  - Added `Roll.CreateBaseline` method and corresponding tests
  - Added `State.CreateBaseline` method for writing `baseline` migrations to `pgroll.migrations`
  - Extended migration type definition to include 'baseline' type

The `pgroll baseline` command is currently hidden. Un-hiding the command will happen when the `pull` and `migrate` commands are made baseline-aware.

This is the first part of #364

### DIFF
--- a/cli-definition.json
+++ b/cli-definition.json
@@ -11,6 +11,25 @@
       "args": []
     },
     {
+      "name": "baseline",
+      "short": "Create a baseline migration for an existing database schema",
+      "use": "baseline <version> <target directory>",
+      "example": "",
+      "flags": [
+        {
+          "name": "json",
+          "shorthand": "j",
+          "description": "output in JSON format instead of YAML",
+          "default": "false"
+        }
+      ],
+      "subcommands": [],
+      "args": [
+        "version",
+        "directory"
+      ]
+    },
+    {
       "name": "complete",
       "short": "Complete an ongoing migration with the operations present in the given file",
       "use": "complete <file>",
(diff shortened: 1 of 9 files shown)


kvch (approved):
The code and the highlevel functionality looks good to me. I have one minor usability improvement ideas for a later PR:
It would be nice if `pgroll` would create the local baseline migration after the baselining was successful. I got an error from `baseline` a few times, and the migration file was there in my migrations folder, even though the database was not baselined.

andrew-farries:
> It would be nice if pgroll would create the local baseline migration after the baselining was successful. I got an error from baseline a few times, and the migration file was there in my migrations folder, even though the database was not baselined.

Resolved in 71459ab1cbf329b498a64aba9478863e3141ebd4. The migration file is cleaned up if creating the baseline in the target database fails.

PR #835, opened 2025-05-15
Make `start` and `migrate` fail on non-empty schema without migration histories

andrew-farries:
Make the `pgroll start` and `pgroll migrate` commands fail when run against a non-empty schema without a migration history. The error message suggests running `pgroll baseline` before re-running `start` or `migrate`.

Running an initial migration against a non-empty schema is not recommended becauses it can result in a migration that can't be rolled back if the migration references pre-existing objects in the schema.

The recommended workflow is:

1. Create the initial baseline migration
```
$ pgroll baseline 00_initial_baseline migrations/
```

2. Run `migrate` as usual

```
$ pgroll migrate migrations/ --complete
```

This ensures that the full schema is brought into `pgroll` before running the first migration.


gulcin:
@andrew-farries could not we run `pgroll baseline` as part of `start` and `migrate` when run against non-empty schema?

andrew-farries:
> @andrew-farries could not we run `pgroll baseline` as part of `start` and `migrate` when run against non-empty schema?

We could automatically run the baseline, but we'd have to prompt the user for a baseline version name as the `baseline` command does. The user may also want to populate and test the placeholder baseline migration before they proceed with applying any other migrations.

It feels more like this is a scenario where the user should be in control of the baseline rather than making it happen automatically.

exekias:
Having it happen after a prompt sounds good to me!

gulcin:
Yes, absolutely. We could make the user aware, and give them the option to add a version name. I think it would still be an improvement if we let the baseline process start and once that's complete, prompted with complete, a bit more hand-holding at that stage too.

andrew-farries:
I made the CTA to run `pgroll baseline` more obvious in 808912d3d5c2d150392c211207ab224a6b75741e.

Not sure about prompting the user directly in `migrate/start`. Currently, the only commands that will wait for user input are:

* `create`
* `baseline` 

`migrate` is going to be run in automated workflows, so I don't think it should block on user input. We could add a flag to abort if a baseline is needed instead of prompting but maybe just the call to run `baseline` in the output is enough here without adding more flags to the `migrate` command?

gulcin:
> The CTA to run `pgroll baseline` more is more obvious in [808912d](https://github.com/xataio/pgroll/commit/808912d3d5c2d150392c211207ab224a6b75741e).
> 
> Not sure about prompting the user directly in `migrate/start`. Currently, the only commands that will wait for user input are:
> 
> * `create`
> * `baseline`
> 
> `migrate` is going to be run in automated workflows, so I don't think it should block on user input. We would need a flag to abort if a baseline is needed instead of prompting but maybe just the call to run `baseline` in the output is enough here without adding more flags to the `migrate` command?

Good point, we would not want to break CI/CD processes of our users. I am okay with  "Run `pgroll baseline` first" in the output.

cmd/start.go
```diff
@@ -40,6 +41,16 @@ func startCmd() *cobra.Command {
 			}
 			defer m.Close()
 
+			// Check whether the schema needs an initial baseline migration
+			needsBaseline, err := m.State().HasExistingSchemaWithoutHistory(ctx, m.Schema())
```
kvch, line 45: Why not pass `cmd.Context()` as a parameter in this function?

kvch (approved):

PR #847, opened 2025-05-20
Mark hidden commands in `build-cli-definition` tool

andrew-farries:
~~Make the `build-cli-definition` tool not include hidden commands in the generated CLI definition JSON file.~~

~~Hidden commands are either internal (like `anaylze`) or work in progress (like `baseline`) and shouldn't be included in the generated CLI definition.~~

Updated based on https://github.com/xataio/pgroll/pull/847#pullrequestreview-2853771576.

Make the `build-cli-definition` tool mark hidden commands in the generated JSON output.

### DIFF
--- a/cli-definition.json
+++ b/cli-definition.json
@@ -5,6 +5,7 @@
       "name": "analyze",
       "short": "Analyze the SQL schema of the target database",
       "use": "analyze",
+      "hidden": true,
       "example": "",
       "flags": [],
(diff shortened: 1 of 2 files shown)


divyenduz (approved):
LGTM. My preferred solution will be to add a hidden: true column though instead of removing. Having `baseline` structure before it is available will help with early integration.

PR #856, opened 2025-05-23
Add new subcommand to validate migration files

### LINKED ISSUE #370: Validate a migration file
Frontend uses the JSON schema to validate pgroll JSON migrations, however, columns like default should be an SQL expression can only be represented as String in the JSON schema. This makes validating them hard. Maybe `pgroll` can provide a validate command/API that can be used to see if a JSON migration is correct or not? Frontend can use this to validate before starting the migration and `pgroll` CLI users too can benefit from it.

As a CLI command, this can be something like: `pgroll validate <migrationfile>`

kvch:
This PR adds a new subcommand to validate migration files.

```
```

It is first step towards providing a way to validate migration files before running them. At the moment it can validate a single migration file against the latest schema in the database. This can be added to your local workflow, or a linter to make sure that the newly added migration file is correct.

Future improvements can be
* validating multiple migrations
* validating all unapplied migrations in a folder
* verifying `up` migrations on a subset of the table

The PR contains some minor changes to the migration phase `Start`. Now migration validation is decoupled from it. It must be called explicitly before running `Start`.

Closes #370

### DIFF
--- a/pkg/roll/execute.go
+++ b/pkg/roll/execute.go
@@ -15,7 +15,23 @@ import (
 	"github.com/xataio/pgroll/pkg/schema"
 )
 
+func (m *Roll) Validate(ctx context.Context, migration *migrations.Migration) error {
+	if m.skipValidation {
+		return nil
+	}
+	lastSchema, err := m.state.LastSchema(ctx, m.schema)
+	if err != nil {
+		return err
+	}
+	err = migration.Validate(ctx, lastSchema)
+	if err != nil {
+		return fmt.Errorf("migration '%s' is invalid: %w", migration.Name, err)
+	}
+	return nil
+}
+
 // Start will apply the required changes to enable supporting the new schema version
+// Migrations must be validated before running Start.
 func (m *Roll) Start(ctx context.Context, migration *migrations.Migration, cfg *backfill.Config) error {
 	// Fail early if we have existing schema without migration history
 	hasExistingSchema, err := m.state.HasExistingSchemaWithoutHistory(ctx, m.schema)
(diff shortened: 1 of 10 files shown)


pkg/state/state.go
```diff
@@ -236,24 +236,29 @@ func (s *State) Status(ctx context.Context, schema string) (*Status, error) {
 // this will effectively activate a new migration period, so `IsActiveMigrationPeriod` will return true
 // until the migration is completed
 // This method will return the current schema (before the migration is applied)
-func (s *State) Start(ctx context.Context, schemaname string, migration *migrations.Migration) (*schema.Schema, error) {
+func (s *State) Start(ctx context.Context, schemaname string, migration *migrations.Migration) error {
```
kvch, line 239: Fetching the last schema version is decoupled from starting a new migration.

pkg/roll/execute.go
```diff
@@ -15,7 +15,23 @@ import (
...
+	}
+	err = migration.Validate(ctx, lastSchema)
+	if err != nil {
+		return fmt.Errorf("migration '%s' is invalid: %w", migration.Name, err)
+	}
+	return nil
+}
+
 // Start will apply the required changes to enable supporting the new schema version
+// Migrations must be validated before running Start.
```
andrew-farries, line 34: How about having this `Start` method also perform validation, even though we have broken validation out into a separate step?

Previously the API of this package was

```
Start
Complete / Rollback
```

Now consumers need to remember to validate first:

```
Validate
Start
Complete / Rollback
```

Not sure there is a use-case where you'd deliberately want to Start without validating first.

WDYT?

andrew-farries (commented):
Nice, left one comment

pkg/roll/execute.go
kvch: Works for me.

andrew-farries (approved):

PR #860, opened 2025-05-28
Set default values for columns in new schema version views

### LINKED ISSUE #42: Views don't respect default values
As demonstrated by @urso, updating items through views doesn't respect the `DEFAULT` keyword:

```
CREATE SCHEMA testing;

SET SEARCH_PATH TO testing;

CREATE TABLE tbl (
  id INT PRIMARY KEY,
  i INT NOT NULL DEFAULT 0
);

CREATE VIEW v AS SELECT * FROM tbl;

-- inserting works. `i` is set to DEFAULT.
INSERT INTO v (id) VALUES (1);

-- Inserting via v with explicit 'DEFAULT' works as well.
INSERT INTO v (id, i) VALUES (2, DEFAULT);

SELECT * FROM v;

--  Error: not-null constraint violated because the DEFAULT on the view is NULL while there is no DEFAULT on tbl.
-- We can not use DEFAULT in ON CONFLICT clauses.
INSERT INTO v (id) VALUES (1) ON CONFLICT (id) DO UPDATE SET i=DEFAULT;

-- Simple update with DEFAULT fails.
UPDATE v SET i=1 WHERE id=2; -- set `i` to 1 to verify output in SELECT below.
UPDATE v SET i=DEFAULT WHERE id=2; -- this fails :(

SELECT * FROM v;

DROP SCHEMA testing CASCADE;
```

We can actually configure default values at the view level by doing an alter view:
```
ALTER VIEW v ALTER COLUMN i SET DEFAULT 0;
```

kvch:
At the moment, if you try to update or insert default values in a view of a new schema version, it fails because the
default value is unset. This PR sets the default value for the new schema version.

Closes #42


kvch:
Coverage job is failing because I opened a PR from my fork.

andrew-farries (approved):

PR #877, opened 2025-06-10
Add more details to `update` subcommand page

kvch:
This PR adds more details to `update` command.

Requires #855

### DIFF
--- a/docs/cli/update.mdx
+++ b/docs/cli/update.mdx
@@ -1,6 +1,6 @@
 ---
 title: Update
-description: Update migration files that contain breaking changes
+description: Update migration files to the latest file format
 ---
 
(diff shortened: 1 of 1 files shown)


andrew-farries (commented):
I think the guide can be much simpler; maybe so much so that it could be part of the doc page for the `update` command itself.

I don't think we need to distinguish the case where you update whatever files you have locally vs pulling first and then updating.

The doc only needs to:
* explain why you'd want to run `pgroll update` (the breaking change case you describe here).
* give an example of how you use it against a local `migrations/` folder.

I'd be in favour of putting that in the docs for the `update` command and not having this guide.

kvch:
Thank you!

Issue #742, opened 2025-03-24
Unify all database actions


kvch:
### Goal

My goal is to unify all database actions pgroll is executing. Example actions include dropping columns, renaming columns, creating and dropping triggers, creating constraints, etc.

We can define a list of these actions in a migration phase and run them sequentially. The following example is a `Complete` phase of an imaginary migration:

```golang
func Complete(ctx context.Context) []DBAction {
    return []DBAction{
        NewDropColumnAction(conn, tableName, columnName),
        NewRenameColumnAction(conn, tableName, newName, oldName),
        NewDropColumnAction(conn, tableName, CNeedsBackfill),
        NewDropTriggerAction(conn, triggerName),
    }
}
```

### TODO
- [ ] define new `DBAction` interface https://github.com/xataio/pgroll/pull/709
- [ ] drop columns https://github.com/xataio/pgroll/pull/709
- [ ] rename column
- [ ] create index
- [ ] create unique index
- [ ] create constraints
- [ ] create trigger
- [ ] drop trigger
- [ ] create function
- [ ] drop function
- [ ] create new `dbaction` package
- [ ] change the interface of migration phases to return a list of DBActions

kvch:
All actions are transformed, and migration phases are adjusted.

PR #880, opened 2025-06-10
Use `DropFunctionAction` in remaning operations

kvch:
Related #742

### DIFF
--- a/pkg/migrations/op_add_column.go
+++ b/pkg/migrations/op_add_column.go
@@ -212,8 +212,7 @@ func (o *OpAddColumn) Rollback(ctx context.Context, l Logger, conn db.DB, s *sch
 		return err
 	}
 
-	_, err = conn.ExecContext(ctx, fmt.Sprintf("DROP FUNCTION IF EXISTS %s CASCADE",
-		pq.QuoteIdentifier(TriggerFunctionName(o.Table, o.Column.Name))))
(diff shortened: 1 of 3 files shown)


pkg/migrations/trigger.go
```diff
@@ -41,6 +41,11 @@ type createTriggerAction struct {
 	cfg  triggerConfig
 }
 
+type dropTriggerAction struct {
```
andrew-farries, line 44: Shouldn't this and the related code go with the other actions in `dbactions.go`?

andrew-farries: Also, this appears to duplicate the existing `DropFunctionAction`

pkg/migrations/trigger.go
```diff
@@ -85,6 +90,19 @@ func (a *createTriggerAction) Execute(ctx context.Context) error {
...
 }
 
+func NewDropTriggerAction(conn db.DB, name string) DBAction {
+	return &dropTriggerAction{
+		conn: conn,
+		name: name,
+	}
+}
+
+func (a *dropTriggerAction) Execute(ctx context.Context) error {
```
andrew-farries, line 100: The action is called `dropTriggerAction` but actually drops a function.

andrew-farries (commented):
Not sure about this one.

andrew-farries (approved):

PR #896, opened 2025-06-13
Update `pgroll latest` to take account of migration `version_schema`

andrew-farries:
#884 added support for the `version_schema` field in migration files, allowing a migration to specify the name of the version schema that will be created when the migration is applied:

```yaml
# migration files can now specify the version schema 
# name to be created by the migration
version_schema: my_version_schema
operations:
  - create_table:
      name: items
      columns:
        - name: id
          type: serial
          pk: true
        - name: name
          type: varchar(255)
```

When applied as `pgroll start 01_create_table.yaml`, this migration will take the name of its version schema from the migration file:

```
+-----------------------+-------------------+
| Name                  | Owner             |
|-----------------------+-------------------|
...
| public_my_version_schema | postgres       |
+-----------------------+-------------------+
```

If `version_schema` is not specified, the name of the version schema defaults to the filename, as before.

---

This PR updates the `pgroll latest` command to take account of this new behaviour, where the version schema name can be decoupled from its filename.

Previous behaviour:

* `pgroll latest schema`: would return the most recent migration name prefixed with the schema name, eg `public_01_create_table`.
* `pgroll latest migration`: would return the most receent migration name without the schema prefix, eg `01_create_table`.

New behaviour:

* `pgroll latest schema`: returns the version schema name of the most  recent migration prefixed with the schema name, eg  `public_01_set_by_version_schema_field`
* `pgroll latest migration`: returns the name of the most recent   migration, without the schema prefix, eg, `01_create_table`.

## Examples

Run:

```
pgroll migrate examples/ --complete
```

to apply all example migrations. Now:

```bash
# get the **name** of the latest migration applied to the target database
$ pgroll latest migration
56_with_version_schema
```

```bash
# get the **version schema name** of the latest migration applied to the target database
$ pgroll latest schema
public_with_version_schema
```

```bash
# get the **name** of the latest migration in the migrations/ folder
$ pgroll latest migration --local examples/
56_with_version_schema
```

```bash
# get the **version schema name** of the latest migration in the migrations/ folder
$ pgroll latest schema --local examples/
public_with_version_schema
```


kvch (approved):

andrew-farries:
> LGTM, as I understand it, this is an internal change and not breaking in the public API correct?

Yes, the command names and flags are unchanged.

PR #897, opened 2025-06-13
New `DBAction` for adding unique constraints using indices

kvch:
Relates to #742

### DIFF
--- a/pkg/migrations/op_add_column.go
+++ b/pkg/migrations/op_add_column.go
@@ -8,8 +8,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/lib/pq"
-
 	"github.com/xataio/pgroll/internal/defaults"
 	"github.com/xataio/pgroll/pkg/backfill"
 	"github.com/xataio/pgroll/pkg/db"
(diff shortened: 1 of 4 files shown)


pkg/migrations/op_add_column.go
```diff
@@ -177,12 +175,7 @@ func (o *OpAddColumn) Complete(ctx context.Context, l Logger, conn db.DB, s *sch
 	}
 
 	if o.Column.Unique {
-		_, err = conn.ExecContext(ctx, fmt.Sprintf("ALTER TABLE %s ADD CONSTRAINT %s_%s_key UNIQUE USING INDEX %s",
-			pq.QuoteIdentifier(o.Table),
-			o.Table,
-			o.Column.Name,
-			UniqueIndexName(o.Column.Name)))
-		if err != nil {
+		if err := NewAddConstraintUsingUniqueIndex(conn, o.Table, o.Column.Name, UniqueIndexName(o.Column.Name)).Execute(ctx); err != nil {
```
andrew-farries, line 178: nit: for a line this long I think it's better to do the assign to `err` on one line and the `if` check on the next.

pkg/migrations/rename.go
```diff
@@ -120,14 +119,11 @@ func (a *renameDuplicatedColumnAction) Execute(ctx context.Context) error {
...
 			// Create a unique constraint using the unique index
-			createUniqueConstraintSQL := fmt.Sprintf(cCreateUniqueConstraintSQL,
-				pq.QuoteIdentifier(a.table.Name),
-				pq.QuoteIdentifier(StripDuplicationPrefix(idx.Name)),
-				pq.QuoteIdentifier(StripDuplicationPrefix(idx.Name)),
-			)
-
-			_, err = a.conn.ExecContext(ctx, createUniqueConstraintSQL)
-			if err != nil {
+			if err := NewAddConstraintUsingUniqueIndex(a.conn,
```
andrew-farries, line 122: agan, would rather split the assign to `err` and the `if check..

Issue #872, opened 2025-06-10
Provide an option to not create version schema


andrew-farries:
Creating a version schema for a migration may be undesirable for a specific migration, or a user may wish to not create version schema for **any** migrations.

We should provide a mechanism to avoid creating version schema when starting migrations.

andrew-farries:
Something to consider here:

* what should `pgroll status` and `pgroll latest` show for migrations that didn't create version schema.

PR #916, opened 2025-06-17
Make `pgroll latest schema` ignore migrations for which no version schema exists

### LINKED ISSUE #915: `pgroll latest schema` incorrectly returns versions for which no version schema exists
The `pgroll latest schema` command fails to consider whether a version schema actually exists for the most recent migration, returning the latest version schema regardless of its existence.

## Reproduction

1. Apply a migration
```
pgroll  start examples/01_create_tables.yaml --complete
```

2. Create an inferred migration
```
CREATE TABLE foo(id int PRIMARY KEY)
```

3. Run `pgroll latest schema`

## Expected behaviour

`pgroll latest schema` should return:

```
public_01_create_tables
```

## Actual behaviour

`pgroll latest schema` returns:

```
public_sql_fca0901e
```

---

Inferred migrations don't create version schema, so the most recent version schema after the `CREATE TABLE...` is still `public_01_create_tables`, not the version schema name of the inferred migration.

The `pgroll latest schema` command should return the most recent version schema name that actually exists in the database.

Related to #872

andrew-farries:
Fix the problem described in #915 where `pgroll latest schema` will return version schema names for which no version schema exists in the database, such as for `inferred` migrations.

---

1. Apply a migration
```
pgroll  start examples/01_create_tables.yaml --complete
```

2. Create an inferred migration
```
CREATE TABLE foo(id int PRIMARY KEY)
```

3. Run `pgroll latest schema`

Before this PR the output would be incorrect:

```
public_sql_fca0901e
```

The most recent migration (inferred from the `CREATE TABLE`) did not create a version schema - the most recent schema version is still `public_01_create_tables`.

This PR corrects the behaviour of `pgroll latest schema` to consider only migrations for which a version schema exists. So the output is:

```
public_01_create_tables
```

as expected.

---

Fixes #915 

Related to #872

### DIFF
--- a/pkg/roll/execute.go
+++ b/pkg/roll/execute.go
@@ -175,7 +175,7 @@ func (m *Roll) Complete(ctx context.Context) error {
 
 	// Drop the old schema
 	if !m.disableVersionSchemas {
-		prevVersion, err := m.state.PreviousVersion(ctx, m.schema, false)
+		prevVersion, err := m.state.PreviousVersion(ctx, m.schema)
 		if err != nil {
 			return fmt.Errorf("unable to get name of previous version: %w", err)
 		}
--- a/pkg/roll/latest_test.go
+++ b/pkg/roll/latest_test.go
@@ -72,7 +72,37 @@ func TestLatestVersionRemote(t *testing.T) {
 			latestVersion, err := m.LatestVersionRemote(ctx)
 			require.NoError(t, err)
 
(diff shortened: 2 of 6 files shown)


andrew-farries:
Coverage is failing because this is stacked on https://github.com/xataio/pgroll/pull/914.

When rebased it should show increased coverage of the `pkg/state` package.

PR #923, opened 2025-06-19
Move `TriggerConfig` to `backfill` package

kvch:
This PR is the first step towards moving creating triggers right before the backfill process starts.

In the long run, operations will only pass a trigger configuration to the `backfill.Backfill` struct, and it will
take care of loading the proper triggers for each table.

### DIFF
--- a/pkg/backfill/trigger.go
+++ b/pkg/backfill/trigger.go
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package backfill
+
+import (
+	"github.com/xataio/pgroll/pkg/schema"
+)
+
+type TriggerDirection string
+
+const (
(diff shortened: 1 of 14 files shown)


andrew-farries (approved):
Changes look fiine 👍 

Can you explain a bit more the motivation behind this:

> first step towards moving creating triggers right before the backfill process starts

is this so that we can consolidate trigger creation and only create one per table?

I thought we might have a migration orchestrator component with a signature roughly like:

```go
type Something interface {
  Process(actions ...DbAction) []DbAction
}
```

which would take care of simplifying and combining multiple create trigger actions on the same table, as well as any other changes to the list of actions required before the actions get executed. Is this roughly where you think this refactoring is heading?

PR #924, opened 2025-06-19
Return `backfill.Task` from `Start` phase operations

kvch:
From now on `Start` functions of each operations return a `backfill.Task` instance. This struct contains information required by the backfill process. The tasks are collected in a single `backfill.Job`.
This aggregates all relevant information for the backfill process: list of tables to backfill and all the triggers to load.

### DIFF
--- a/pkg/backfill/backfill.go
+++ b/pkg/backfill/backfill.go
@@ -21,8 +21,8 @@ const CNeedsBackfillColumn = "_pgroll_needs_backfill"
 
 // Task represents a backfill task for a specific table from an operation.
 type Task struct {
-	Table    *schema.Table
-	Triggers []TriggerConfig
+	table    *schema.Table
+	triggers []TriggerConfig
 }
 
 // Job is a collection of all tables that need to be backfilled and their associated triggers.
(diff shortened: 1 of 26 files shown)


andrew-farries (commented):
Some comments and a question about how this relates to the `DBAction` refactoring.

AIUI, the `DBAction` refactoring was going to make the operation `Start` methods return `[]DBAction` which would then be merged/reordered/combined by a new orchestrator component. This component could also include combining triggers that are intended to act on the same table.

With this PR, we make operation `Start` methods return backfill jobs, which will allow us to combine backfill triggers, but what happens to the idea to return `[]DBAction` from `Start`? It seems like returning `[]DBAction` would allow us to solve a greater class of problems, including the one addressed by this PR.

How do you see the changes here fitting in with the broader plan to return `[]DBAction` from operation `Start` methods?

andrew-farries (approved):
Thanks for making the suggested changes 👍 

Just to understand the future direction, Is the plan now to make `Start` return `[]DBAction` in addition to `backfill.Task`?

kvch:
> Just to understand the future direction, Is the plan now to make Start return `[]DBAction` in addition to `backfill.Task`?

Yes. However, I am unhappy with returning 3 things, so I will try to find a way to decrease this number.

PR #929, opened 2025-06-23
Return `DBActions` from `Rollback`

kvch:
This PR changes the signature of `Rollback`. From now on, it returns a list of `DBActions`.


pkg/migrations/op_add_column.go
```diff
@@ -206,33 +206,23 @@ func (o *OpAddColumn) Complete(ctx context.Context, l Logger, conn db.DB, s *sch
...
 }
 
-func (o *OpAddColumn) Rollback(ctx context.Context, l Logger, conn db.DB, s *schema.Schema) error {
+func (o *OpAddColumn) Rollback(l Logger, conn db.DB, s *schema.Schema) ([]DBAction, error) {
 	l.LogOperationRollback(o)
 
 	table := s.GetTable(o.Table)
 	if table == nil {
-		return TableDoesNotExistError{Name: o.Table}
+		return []DBAction{}, TableDoesNotExistError{Name: o.Table}
```
andrew-farries, line 214: can we return `nil` slices rather than empty slices?

The same also applies to the error returns in the `Complete` functions.

andrew-farries (commented):
Would prefer to return `nil` slices here and in `Complete` if possible.

andrew-farries (approved):

PR #930, opened 2025-06-23
Support altering primary key columns that are not included in FK constraints

### LINKED ISSUE #888: Preserve primary key constraint on altered columns
If a column with a primary key constraint on it is altered by `pgroll`, the primary key constraint is lost after completing the operation.

Migration to reproduce the issue:

```yaml
operations:
- alter_column:
    column: myid
    down: "CAST(\"myid\" AS bigint)"
    table: my_table
    type: uuid
    up: "gen_random_uuid"
```

Broken schema:
```
# \d my_table;
                    Table "public.my_table"
  Column   |           Type           | Collation | Nullable | Default
-----------+--------------------------+-----------+----------+---------
 myid      | uuid                     |           | not null |
Indexes:
    "my_table_pkey" UNIQUE, btree ("myid")
```

The primary key constraint is missing.

### Workaround

Run a second migration to apply primary key constraint again:
```yaml
operations:
- create_constraint:
    name: my_table_pk
    columns: [myid]
    table: my_table
    up:
      myid: myid
    down:
      myid: myid
```

kvch:
In the past `pgroll` could not preserve primary key constraint on altered columns. Instead the primary key constraint was down-graded to a unique, not null constraint. Now the PK constraint is not removed during completing the migration.

Futhermore, if the table had rows, `pgroll` could not backfill rows. `pgroll` "updates" the primary keys of the backfilled table to trigger backfilling existing columns. Previously, finding the identity columns returned the duplicated column `__pgroll_new_id` instead of the correct `id`. In this case we need the current PK, so we can copy the correct values to the new column.

I opened follow-up issue to support altering mutiple columns of the same composite key in the same migration: https://github.com/xataio/pgroll/pull/930

Support for alter PK columns that are part of FK constraints is in a follow-up PR: https://github.com/xataio/pgroll/pull/935

Closes #888

### DIFF
--- a/internal/testutils/error_codes.go
+++ b/internal/testutils/error_codes.go
@@ -3,11 +3,12 @@
 package testutils
 
 const (
-	CheckViolationErrorCode     string = "check_violation"
-	ExclusionViolationErrorCode string = "exclusion_violation"
-	FKViolationErrorCode        string = "foreign_key_violation"
-	NotNullViolationErrorCode   string = "not_null_violation"
-	UndefinedColumnErrorCode    string = "undefined_column"
-	UndefinedTableErrorCode     string = "undefined_table"
-	UniqueViolationErrorCode    string = "unique_violation"
+	CheckViolationErrorCode         string = "check_violation"
+	ExclusionViolationErrorCode     string = "exclusion_violation"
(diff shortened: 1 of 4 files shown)


andrew-farries (approved):
Looks good 👍

PR #931, opened 2025-06-23
Add trigger configuration to `backfill.Task`

kvch:
This PR introduces consolidates backfilling of the same column into a single trigger. I am merging all triggers for a table into a single trigger in a follow-up PR.

I added a new data structure `OperationTrigger` that contains the trigger configuration of an operation. Trigger consolidation happens when a backfill task is added to the job. `triggerConfig` stayed the consolidated configuration storage for the trigger template.

In a later PR I am removing `latestSchema` from the parameter list of `Start`, as it is no longer necessary. We can pass it from `Roll` to the `backfill.Job`: https://github.com/xataio/pgroll/pull/934


andrew-farries (commented):
I think we need to add some integration tests here, to check how operations on the same column behave after this change.

For example, the following pair of migrations don't work currently:

```yaml
operations:
  - create_table:
      name: users
      columns:
        - name: id
          type: serial
          pk: true
        - name: email
          type: varchar(255)
          nullable: true
        - name: name
          type: varchar(100)
          nullable: false
        - name: created_at
          type: timestamp
          default: now()
```

```yaml
operations:
  - alter_column:
      table: users
      column: email
      check:
        name: email_format
        constraint: "email IS NULL OR email LIKE 'user%'"
      up: "CASE WHEN email IS NULL THEN email ELSE REPLACE(email, '.com', '.xyz') END"
      down: email
  - alter_column:
      table: users
      column: email
      nullable: false
      up: "COALESCE(email, 'required@example.com')"
      down: email
```

The second migration contains two operations on the `email` field that both need to create a trigger, but the migration fails at the column duplication stage.

kvch:
The test you shared is indeed not working, it is not working on `main` either. It is not supposed to work yet, as the feature is still in progress. The problem is not with the triggers. The root cause of the issue is that the column duplicator tries to duplicate the column multiple times. The PR for addressing this problem is coming in a follow-up PR.

pkg/backfill/backfill.go
```diff
@@ -38,16 +42,62 @@ type Backfill struct {
...
+	return strings.ReplaceAll(sql, from, fmt.Sprintf("NEW.%s", pq.QuoteIdentifier(to)))
+}
+
+func findColumnName(columns map[string]*schema.Column, columnName string) string {
+	for name, col := range columns {
+		if col.Name == columnName {
+			return name
+		}
+	}
+	return columnName
```
andrew-farries, line 100: Can we add comments for this pair of functions. It's not obvious to me why they are now necessary.

pkg/backfill/templates/function.go
```diff
@@ -20,7 +20,9 @@ const Function = `CREATE OR REPLACE FUNCTION {{ .Name | qi }}()
         FROM current_setting('search_path');
 
       IF search_path {{- if eq .Direction "up" }} != {{- else }} = {{- end }} {{ .LatestSchema | ql }} THEN
-        NEW.{{ .PhysicalColumn | qi  }} = {{ .SQL }};
+      {{- $physicalColumn := .PhysicalColumn | qi  }}{{ range $s := .SQL }}
+        NEW.{{ $physicalColumn  }} = {{ $s }};
+	  {{- end }}
```
andrew-farries, line 25: nit: indentation

pkg/backfill/backfill.go
```diff
@@ -38,16 +42,62 @@ type Backfill struct {
...
+		} else {
+			j.triggers[trigger.Name] = triggerConfig{
+				Name:                trigger.Name,
+				Direction:           trigger.Direction,
+				Columns:             trigger.Columns,
+				SchemaName:          j.schemaName,
+				TableName:           trigger.TableName,
+				PhysicalColumn:      trigger.PhysicalColumn,
+				LatestSchema:        j.latestSchema,
+				SQL:                 []string{trigger.SQL},
```
andrew-farries, line 83: if we are rewriting trigger SQL using `rewriteTriggerSQL` when we combine two triggers why is rewrite not needed here?

pkg/backfill/backfill.go
```diff
@@ -38,16 +42,62 @@ type Backfill struct {
...
 func (j *Job) AddTask(t *Task) {
-	j.Tables = append(j.Tables, t.table)
-	j.triggers = append(j.triggers, t.triggers...)
+	if t.table != nil {
+		j.Tables = append(j.Tables, t.table)
+	}
+
+	for _, trigger := range t.triggers {
+		if tg, exists := j.triggers[trigger.Name]; exists {
+			tg.SQL = append(tg.SQL, rewriteTriggerSQL(trigger.SQL, findColumnName(tg.Columns, tg.PhysicalColumn), tg.PhysicalColumn))
```
andrew-farries, line 72: Why do we need to rewrite trigger SQL here?

pkg/backfill/backfill.go
andrew-farries: Thanks for the explanation + comments 👍

PR #935, opened 2025-06-25
Allow altering PK columns even if they are referenced

### LINKED ISSUE #375: Noob migration question
I'm trying to change a column's `id` column from `integer` to `bigint`, but I'm struggling to complete the migration because other tables reference `id` as a foreign key.

I tried following this example: https://github.com/xataio/pgroll/blob/main/docs/README.md#change-type

Here's what I have:

```json
{
  "name": "00_testing",
  "operations": [
    {
      "alter_column": {
        "table": "inspection_item",
        "column": "id",
        "type": "bigint",
        "up": "CAST(id as bigint)",
        "down": "CAST(id as integer)"
      }
    }
  ]
}
```

When I try to `complete` the migration, I get this error:

image

I see examples for adding/deleting foreign key constraints, but I don't see any examples of changing the type of a column that is referenced as a foreign key.

Does anyone have examples of this?

kvch:
Previously, `pgroll` could not alter PK columns if they were included in a foreign key reference. The old column
could not be dropped because it was still referred by foreign key constraints. Now the foreign key constraints are
recreated so they refer to the new PK column.

Requires #930
Closes #375

### DIFF
--- a/pkg/migrations/dbactions.go
+++ b/pkg/migrations/dbactions.go
@@ -946,12 +946,12 @@ func (a *setReplicaIdentityAction) Execute(ctx context.Context) error {
 type alterReferencesAction struct {
 	conn         db.DB
 	id           string
-	referencedBy schema.ReferencedBy
+	referencedBy map[string][]*schema.ReferencedBy
 	table        string
 	column       string
 }
 
-func NewAlterReferencesAction(conn db.DB, referencedBy schema.ReferencedBy, table, column string) *alterReferencesAction {
+func NewAlterReferencesAction(conn db.DB, referencedBy map[string][]*schema.ReferencedBy, table, column string) *alterReferencesAction {
 	return &alterReferencesAction{
 		conn:         conn,
 		id:           fmt.Sprintf("alter_references_%s_%s", table, column),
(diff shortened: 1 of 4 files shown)


pkg/migrations/dbactions.go
```diff
@@ -840,3 +840,77 @@ func (a *setReplicaIdentityAction) Execute(ctx context.Context) error {
...
+	table string
+	def   string
+}
+
+func (a *alterReferencesAction) constraintDefinitions(ctx context.Context) ([]constraintDefinition, error) {
+	rows, err := a.conn.QueryContext(ctx, fmt.Sprintf(`
+SELECT conname, r.conrelid::regclass, pg_catalog.pg_get_constraintdef(r.oid, true) as condef
+FROM pg_catalog.pg_constraint r
+WHERE confrelid = %s::regclass AND r.contype = 'f'`,
+		pq.QuoteIdentifier(a.table),
```
andrew-farries, line 899: If we need this information, I think it should be present in the schema returned by `read_schema`. If we are missing information there we should add it to `read_schema` rather than running ad-hoc queries during the migration.

kvch:
Depends on https://github.com/xataio/pgroll/pull/946

PR #940, opened 2025-06-27
Add simple coordinator for executing `DBActions`

kvch:
This PR adds a coordinator to migrations package. This component is responsible
for executing all registered actions only once in the correct order.

I integrated it already into `Roll` for the complete phase of the migrations.

### DIFF
--- a/pkg/migrations/coordinator_test.go
+++ b/pkg/migrations/coordinator_test.go
@@ -77,10 +77,10 @@ func TestCoordinator(t *testing.T) {
 		"drop default value on column": {
 			actions: []DBAction{
 				NewCreateTableAction(nil, "test_table", "", ""),
-				NewAddColumnAction(nil, "column1", Column{Default: ptr("default_value")}, false),
+				NewAddColumnAction(nil, "test_table1", Column{Name: "column1", Default: ptr("default_value")}, false),
 				NewDropDefaultValueAction(nil, "test_table", "column1"),
 			},
-			expectedOrder: []string{"create_table_test_table", "add_column_column1_", "drop_default_test_table_column1"},
+			expectedOrder: []string{"create_table_test_table", "add_column_test_table1_column1", "drop_default_test_table_column1"},
 		},
 		"alter column multiple times rollback phase": {
 			actions: []DBAction{


pkg/migrations/coordinator.go
```diff
@@ -0,0 +1,61 @@
...
+func moveIdxToLast(actions []string, idx int) []string {
+	if idx < 0 || idx >= len(actions) {
+		return actions
+	}
+	duplicate := actions[idx]
+	actions = append(actions[:idx], actions[idx+1:]...)
+	if len(actions) > 0 && actions[len(actions)-1] == duplicate {
+		// If the last action is the same as the one we want to move, no need to change order
+		return actions
+	}
```
andrew-farries, line 58: All test cases still pass when I remove this check. Missing coverage or redundant check?

AIUI this tries to ensure that any duplicate actions occur at or near the end of the actions list; I think iterating over the slice first to last and simply removing duplicates gets us to the same result?

andrew-farries: An alternative design might be to have the coordinator apply rules, in order, to the actions list.

The only rule we'd need at this stage is a `removeDuplicateActions` rule AIUI.

pkg/roll/execute.go
```diff
@@ -109,16 +109,15 @@ func (m *Roll) StartDDLOperations(ctx context.Context, migration *migrations.Mig
...
-		for _, action := range startOp.Actions {
-			if err := action.Execute(ctx); err != nil {
-				errRollback := m.Rollback(ctx)
-				if errRollback != nil {
-					return nil, errors.Join(
-						fmt.Errorf("unable to execute start operation of %q: %w", migration.Name, err),
-						fmt.Errorf("unable to roll back failed operation: %w", errRollback))
-				}
-				return nil, fmt.Errorf("failed to start %q migration, changes rolled back: %w", migration.Name, err)
+		coordinator := migrations.NewCoordinator(startOp.Actions)
```
kvch, line 112: I would like to move the coordinator one step later, so it collects all actions for all operations in a follow-up PR.

pkg/migrations/coordinator.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		orderedActions = append(orderedActions, action.ID())
+	}
+	return &Coordinator{
+		actions:        actionsMap,
+		orderedActions: orderedActions,
+	}
+}
+
+// Execute runs all actions in the orderedActions they were added to the coordinator.
```
andrew-farries, line 36:
```suggestion
// Execute runs all actions in the order they were added to the coordinator.
```

pkg/migrations/coordinator_test.go
```diff
@@ -0,0 +1,163 @@
...
+			expectedOrder: []string{
+				"drop_column_t1_column3",
+				"rename_duplicated_t1_column3",
+				"drop_column_t1_column1",
+				"rename_duplicated_t1_column1",
+				"drop_column_t1_column2",
+				"rename_duplicated_t1_column2",
+			},
+		},
+		"add same column multiple times to same column": {
```
andrew-farries, line 60:
```suggestion
		"add same column multiple times to same table": {
```

pkg/migrations/coordinator_test.go
```diff
@@ -0,0 +1,163 @@
...
+				NewCreateTableAction(nil, "test_table", "id INT", ""),
+				NewCommentTableAction(nil, "test_table", ptr("This is a test table")),
+				NewCommentColumnAction(nil, "test_table", "id", ptr("This is a test column")),
+			},
+			expectedOrder: []string{"create_table_test_table", "comment_table_test_table", "comment_column_test_table_id"},
+		},
+		"drop default value on column": {
+			actions: []DBAction{
+				NewCreateTableAction(nil, "test_table", "", ""),
+				NewAddColumnAction(nil, "column1", Column{Default: ptr("default_value")}, false),
```
andrew-farries, line 80: This action looks odd compared to how the others are used; the second argument is meant to be the table name, and the `Column` in the third argument should have a name.

pkg/roll/execute.go
andrew-farries: what is the issue that prevents doing us doing so at the moment?

pkg/roll/execute.go
```diff
@@ -212,28 +211,24 @@ func (m *Roll) Complete(ctx context.Context) error {
 
 	// execute operations
 	refreshViews := false
+	var actions []migrations.DBAction
 	for _, op := range migration.Operations {
-		actions, err := op.Complete(m.logger, m.pgConn, currentSchema)
+		opActions, err := op.Complete(m.logger, m.pgConn, currentSchema)
```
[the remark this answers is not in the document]
andrew-farries: This is the what we are aiming for with `Start` and `Rollback` too right?

andrew-farries (commented):
Does. this PR implement the changes required to update the virtual schema between operations in the `Complete` phase that we talked about last week or is that something to address separately?

andrew-farries (approved):

PR #944, opened 2025-07-01
Return list of `DBAction` from `Start`

kvch:
This PR returns adds a new return value to `Start` to return the list of `DBActions`. However, returning 3 values is too many, so in a follow-up PR I am combining the list of `DBActions` and `backfill.Task` into the same data structure. The name is pending, my ideas so far `StartStep` or `StartOperation`. If you have a better name, I am open to using them.

New data structure in a follow-up PR:
```golang
type StartOperation struct {
    Actions []DBAction
    Task *backfill.Task
}
```

### DIFF
--- a/pkg/migrations/op_alter_column.go
+++ b/pkg/migrations/op_alter_column.go
@@ -18,16 +18,16 @@ var (
 	_ Createable = (*OpAlterColumn)(nil)
 )
 
-func (o *OpAlterColumn) Start(ctx context.Context, l Logger, conn db.DB, s *schema.Schema) (*backfill.Task, error) {
+func (o *OpAlterColumn) Start(ctx context.Context, l Logger, conn db.DB, s *schema.Schema) ([]DBAction, *backfill.Task, error) {
 	l.LogOperationStart(o)
 
 	table := s.GetTable(o.Table)
 	if table == nil {
-		return nil, TableDoesNotExistError{Name: o.Table}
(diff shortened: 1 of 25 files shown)


pkg/migrations/op_alter_column.go
```diff
@@ -79,16 +79,19 @@ func (o *OpAlterColumn) Start(ctx context.Context, l Logger, conn db.DB, s *sche
 		},
 	)
 	task := backfill.NewTask(table, triggers...)
+
+	dbActions := make([]DBAction, 0)
```
andrew-farries, line 83:
```suggestion
	var dbActions []DBAction
```

pkg/roll/execute.go
```diff
@@ -101,15 +101,20 @@ func (m *Roll) StartDDLOperations(ctx context.Context, migration *migrations.Mig
 	// execute operations
 	job := backfill.NewJob(m.schema, versionSchemaName)
 	for _, op := range migration.Operations {
-		task, err := op.Start(ctx, m.logger, m.pgConn, newSchema)
+		actions, task, err := op.Start(ctx, m.logger, m.pgConn, newSchema)
```
andrew-farries, line 104: Rename `task` to something more specific like `backfillTask`.

andrew-farries (approved):
Looks good, just a couple of comments 👍

PR #946, opened 2025-07-02
Make column duplication idempotent

kvch:
This PR makes column duplication idempotent. Previously, if the same column was altered in multiple operations in the same migration was not working. The modified column could not be duplicated in the second operation because the `ALTER TABLE` statement has failed. I added `IF NOT EXISTS` to make the column creation idempotent. Now the column is duplicated once, and other changes are added on top of it like setting not null constraints, changing default values, etc.

The only edge case this PR does not address is when the column type is changed in a later operation. This limitation can be worked around by changing the type first in an `alter_column` operation and then adding other changes to the migration. I do opted for this trade-off because it would increase the complexity of the duplicator but not much of an upside.

This is working:

```yaml
operations:
 - alter_column:
    name: mycol
    type: text
 - alter_column:
    name: mycol
    comment: nocomment
```

This is not supported:

```yaml
operations:
 - alter_column:
    name: mycol
    comment: nocomment
 - alter_column:
    name: mycol
    type: text
```

### DIFF
--- a/pkg/migrations/coordinator.go
+++ b/pkg/migrations/coordinator.go
@@ -33,7 +33,7 @@ func NewCoordinator(actions []DBAction) *Coordinator {
 	}
 }
 
-// Execute runs all actions in the orderedActions they were added to the coordinator.
+// Execute runs all actions in the order they were added to the coordinator.
 func (c *Coordinator) Execute(ctx context.Context) error {
 	for _, id := range c.orderedActions {
 		action, exists := c.actions[id]


pkg/migrations/op_alter_column.go
```diff
@@ -63,8 +63,13 @@ func (o *OpAlterColumn) Start(ctx context.Context, l Logger, conn db.DB, s *sche
 	// a variable for the new column. Save the old column name for use as the
 	// physical column name. in the down trigger first.
 	oldPhysicalColumn := column.Name
+	columnType := column.Type
```
kvch, line 66: The column type was missing from the schema and it broke column duplication.

PR #1015, opened 2026-01-17
fix: close connections in `getRowCount` during backfill (connection leak)

JoeyEamigh:
Hi again! This fixes a database connection leak in `getRowCount()` that was causing "too many clients" during an initial table creation migration.

My migration created 142 tables then subsequently attempted to create a bunch of constraints on those tables, triggering a backfill. Because the tables were empty it ran all three of the queries in `getRowCount()`. I'm not very good at golang, but from my understanding the `defer rows.Close()` call only closes the database connection for the first query, leaking one connection for tables with rows and two connections for empty tables. It seems that the `defer` call binds to the first pointer in `rows` and doesn't update when `rows` is reassigned.

If `rows.Close()` is an intensive operation another option would be to defer each of the `rows.Close()` calls in this function. Let me know if that is the preferred approach.

Thanks again!


JoeyEamigh:
fixed by 0a64836

PR #1027, opened 2026-04-15
Ignore per-session temp schemas in inferred migrations

JoeyEamigh:
The `pg_roll_handle_ddl` event trigger in `pkg/state/init.sql` records an inferred row in `pgroll.migrations` for any DDL that happens outside a pgroll-owned connection. It filters pgroll's own sessions via `pgroll.no_inferred_migrations`, but it doesn't exclude PostgreSQL temp schemas.

I have been running pgroll in prod for ~2 months and have 291 inferred rows on `pg_temp` schemas. It's not harmful by any stretch, but is messy.

This PR adds a name-prefix guard in `placeholder.raw_migration()` that filters temp tables and a test.

### DIFF
--- a/pkg/state/init.sql
+++ b/pkg/state/init.sql
@@ -58,6 +58,14 @@ BEGIN
     IF schemaname IS NULL THEN
         RETURN;
     END IF;
+    -- Ignore DDL on PostgreSQL's per-session temporary schemas. Applications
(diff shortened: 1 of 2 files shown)


exekias (approved):
Thank you for contributing! this change makes sense to me, WDYT @andrew-farries ?

PR #1034, opened 2026-05-25
docs(readme): add pgfence to References

flvmnt:
Adds pgfence to the References section. Static analyzer for Postgres migration SQL covering lock modes, blocking DDL, and safe expand/contract rewrites. Complementary to migration runners like pgroll: pgfence flags risks before execution.

Site: https://pgfence.com


exekias:
Thank you for opening the PR, pgfence looks really interesting, I like the approach of detecting possible situations coming from any source, this is an idea we have also played with in the past, as it removes friction when integrating with any other tool, keep it up!

That said, I'm closing this PR without merging because we didn't really use pgfence as a reference when we created pgroll 🙏

PR #1036, opened 2026-05-26
fixed  backfill for generating always identity columns

### LINKED ISSUE #1035: GENERATED ALWAYS columns not supported for operation with backfill ?
Hello,

## Presentation

There seems to be an issue around changes that match two conditions: a **backfill** is required and the tables has "**GENERATED ALWAYS**" column(s).
NB: I'm a beginner in pgroll, so I can't say _for sure_ the conditions for it to happen, above is more an hypothesis than a definitive analysis. But I got a clear & simple reproduction case (cf below).

```
Failed to start migration: unable to backfill table "my_table": pq: column "id" can only be updated to DEFAULT
```

## How to reproduce ?
1. **prepare a new database from scratch** (outside of pgroll) with a simple table:
   ```
   createdb reproduct

   psql reproduct

   create table my_table (
      id        integer GENERATED ALWAYS AS IDENTITY not null primary key,
      name      varchar not null,
      parent_id integer not null
   );

   INSERT INTO my_table(name, parent_id) VALUES ('alice', 100), ('bob', 100);
   ```
2. **install pgroll & import** (macos here):
   ```
   brew install pgroll
   export PGROLL_PG_URL="postgres://postgres@localhost/reproduct?sslmode=disable"

   pgroll init
   SUCCESS  Initialization complete

   pgroll baseline 000_imported_schema ./migrations/
   SUCCESS  Baseline created successfully. Placeholder migration "migrations/000_imported_schema.yaml" written
   ```
3. **test a first basic migration:**
   ```
   cat migrations/001_test.yaml
   operations:
     - alter_column:
         table: my_table
         column: parent_id
         type: smallint
         up: parent_id
         down: parent_id

   pgroll start migrations/001_test.yaml
   ```

## Results

Current (buggy) result: an error ❌
```
$ pgroll start migrations/001_test.yaml
  ERROR   Failed to start migration: unable to backfill table "my_table": pq: column "id" can only be updated to DEFAULT
Error: unable to backfill table "my_table": pq: column "id" can only be updated to DEFAULT
```

Expected result: the migration should "just work", so something like: ✅ 
```
$ pgroll start migrations/001_test.yaml
 SUCCESS  New version of the schema available under the postgres "public_001_test" schema
```

## Notes
I tested quickly variants to "scope" the issue:
- A) If I execute the same scenario with `GENERATED BY DEFAULT` instead of `GENERATED ALWAYS` => no issue to start migration
- B) If I execute the same scenario with an `add_column` operation instead of an `alter_column` => no issue to start migration

And here are some context:
```
pgroll --version
pgroll version 0.16.2
```

```
psql -c "SELECT version();"
PostgreSQL 18.4 (Homebrew) on aarch64-apple-darwin24.6.0, compiled by Apple clang version 17.0.0 (clang-1700.6.4.2), 64-bit
```

## Questions
- Is it expected (ie: a known limitation) ? I haven't seen it
- Is it a use-case that the tool should support ? I would think it's legitimate (since GENERATED ALWAYS is a native & fully supported PG feature)

Thanks, have a good day !

codeafridi:
fixes #1035 

The backfill query used to update the primary key column to itself to fire triggers. PostgreSQL rejects that for `GENERATED ALWAYS` identity columns, so this now updates pgroll’s internal `_pgroll_needs_backfill` column instead.

## Tests

- `go test ./pkg/backfill/...`
- `go test ./pkg/migrations -run 'TestAlterColumnValidation/backfill_with_generated_always_identity_primary_key|TestAlterColumnValidation/backfill_with_multiple_primary_keys'`


codeafridi:
please review @exekias

PR #1047, opened 2026-09-03
Accept the legacy object-shaped columns in create_index

andrewkroh:
OpCreateIndex.Columns changed from an object keyed by column name to an array of IndexField after #697. A migration that called Start under the older shape and was never completed still holds that object shape in the migration store. Decoding it now fails, because the struct field is []IndexField, so Complete can never run and the migration stays stuck.

Give OpCreateIndex its own UnmarshalJSON. It first decodes into the current array shape. If that fails, it decodes into the object shape instead and converts each entry to an IndexField, sorted by column name for a deterministic order.

Encoding is unchanged. MarshalJSON still writes the array shape, so a completed migration is written back in the current format.


exekias (approved):

andrewkroh:
You must run `complete` (pgroll.com/docs/latest/cli/complete) before upgrading to 0.16, which had the breaking change. Assuming you do that, then you don't need this code change.

